PR #39, opened 2018-03-02
added options add-pid-id,add-programinfo-id,add-programinfo-id

haraldjordan78:
Sorry for the mass of changes, i guess lots of them come from visual studio formatting! Please check and tell me whats wrong before i unneccesarily merge my changes for the mux plugin as well :-)

### DIFF
--- a/src/tsplugins/tsplugin_pmt.cpp
+++ b/src/tsplugins/tsplugin_pmt.cpp
@@ -47,7 +47,7 @@ TSDUCK_SOURCE;
 //----------------------------------------------------------------------------
 
 namespace ts {
-    class PMTPlugin: public ProcessorPlugin, private TableHandlerInterface
+    class PMTPlugin : public ProcessorPlugin, private TableHandlerInterface
     {
     public:
         // Implementation of plugin API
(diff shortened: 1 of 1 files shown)


haraldjordan78:
hmmm if i was you i would not be able to determine whats really changed or not. In theory i only added some addional commands and the code for it, without touching existing code. In case you are not able to merge, i would probably need to extract my changes and insert them into your original files manually... what do you think?

lelegard:
Hi Harald,

Indeed, not only for TSDuck but for software engineering in general, it is important to preserve the clarity of the history. This means that the code differences in each commit shall be limited to "real" changes and shall not be polluted by IDE reformatting. Otherwise, not only the coding style of the project becomes inconsistent but it is also very difficult to track what was really changed. This is very important when tracking difficult bugs or regressions.

This is why automatic reformatting of existing code shall be disabled in IDE's. Or at least be limited to code lines which are actually typed. But not to other code.

The command `diff -w` manually detects the real changes (option `-w` means "ignore all forms of spaces"). Using this, I can see that you only added (many) code lines but you did not change existing code, which is good. To manually fix the code, you should keep the output of `diff -w`, revert to the previous version of the file and then manually apply the real changes as kept in `diff -w` output. Of course, this shall be done with IDE reformatting disabled (or using a simpler text editor).

I cannot afford to manually reformat all proposed pull requests. I invest a lot of time in TSDuck, in addition to my other activities. I hope that TSDuck saves a lot of time (and maybe money) to TV professionals and hobbyists. And I do appreciate contributions in return. But I want to preserve the stability and integrity of TSDuck to continue to offer a stable level of quality to TSDuck users.

So, would it be possible for you to reformat your code in the pull request for the `pmt` plugin and use the same precautions before submitting the next pull request ?

Thanks for your cooperation.

haraldjordan78:
Hey Theirry, 
no worries, i shall extract my changes and create a Version thats hopefully worthy to be checked-in. Even if there is no need for this to fulfill the customer requirements of my project, i think this is the best way we can pay back the efforts you put into this which also saved some time in our project. 
Currently i am travelling, i should have a chance to do this on friday. 
I'll close this request and open a new one.

PR #40, opened 2018-03-09
tsplugin_pmt added --add-pid-id --add-programinfo-id

haraldjordan78:
Dear Thierry, 
sorry for the extra-efforts for you reviewing the same change multiple times, please excuse, i am totally new to github. 
There are 2 things in this commit that i am not 100% certain of:

1)
_add_descs(0) - this line does not even exist in your master branch, it is actually _add_descs()

With add_descs(0), it would not compile for me because there is no such constructor overload.

2) 
i am not sure why it shows a removed } in line 588 without a replacement. 

Let me know if i can do any better, i am happy to reingeneer my changes and learn to work with github as long as you are happy to talk to me ;-)

### DIFF
--- a/src/tsplugins/tsplugin_pmt.cpp
+++ b/src/tsplugins/tsplugin_pmt.cpp
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
 //----------------------------------------------------------------------------
 //
 // TSDuck - The MPEG Transport Stream Toolkit
@@ -68,6 +69,18 @@ namespace ts {
             {
             }
         };
+        
+        struct NewStreamDescriptor {
+             uint8_t     descriptor;
+             uint8_t     tag;
+
+             // Constructor.
+             NewStreamDescriptor(uint8_t desc_ = 0, uint8_t tag_ = 0) :
+                 descriptor(desc_),
+                 tag(tag_)
+             {
+             }
+         };
 
         // PMTPlugin instance fields
         bool                _abort;             // Error (service not found, etc)
@@ -93,6 +106,8 @@ namespace ts {
         AudioLanguageOptionsVector _languages;  // Audio languages to set
         SectionDemux        _demux;             // Section demux
         CyclingPacketizer   _pzer;              // Packetizer for modified PMT
+        std::vector<NewStreamDescriptor>    _added_pid_id;      // List of identifiers/descriptors for PID to add
+        uint32_t                            _added_programinfo_id;// Adds a programinfo section to PMT with specified identifier
 
         // Invoked by the demux when a complete table is available.
         virtual void handleTable (SectionDemux&, const BinaryTable&) override;
@@ -120,6 +135,8 @@ ts::PMTPlugin::PMTPlugin(TSP* tsp_) :
     _removed_pid(),
     _removed_desc(),
     _added_pid(),
+    _added_pid_id(),
+    _added_programinfo_id(),
     _moved_pid(),
     _set_servid(false),
     _new_servid(0),
@@ -133,7 +150,7 @@ ts::PMTPlugin::PMTPlugin(TSP* tsp_) :
     _ac3_atsc2dvb(false),
     _eac3_atsc2dvb(false),
     _cleanup_priv_desc(false),
-    _add_descs(0),
+    _add_descs(),
     _languages(),
     _demux(this),
     _pzer()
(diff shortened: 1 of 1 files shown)


lelegard:
Hi Harald,

Thank you for your contribution. I merged it into the master branch. I added your name in the list of contributors (if you don't mind).

I have made a change in your option. I must admit I did not like much the `--add-pid-id` option because, as you pointed out yourself in your help text, this restricted the number of `--add-pid` to one. Additionally, it was designed to work only on new PID's and could not set a component tag or cue type on an existing PID.

So, I kept the whole feature but I changed the option. I replaced `--add-pid-id` with two options `--set-stream-identifier pid/id` and `--set-cue-type pid/type`. Each option apply to a specified PID, whether it is a new one or an existing one. We may continue to use many `--add-pid` and also use many `--set-stream-identifier` and  many `--set-cue-type`.

I also took that opportunity to do some internal refactoring in the plugin. I wrote it about 10 years ago and did not modify it for years, so it was time for a lifting. Specifically, I moved the processing of the two new options in a template method so that we may add several similar types of options without writing code, just calling the template method.

Thanks again for your contribution and have a nice week-end.
Thierry

haraldjordan78:
Hey Thierry, 

thats great news, thanks a lot! Next Thing to come are my modifications to the mux plugin, hope you like them…

Sure i am happy for you listing me in the contributor list, thanks for that. 
Also i totally agree with the renamed Options you did, it makes much more sense that way.
The template you added is a very smart Thing. I have to admit, you are coding on a totally different Level than me. So i have to excuse for the Basic coding level of my additions.

What does not make sense in first place for me is the bare fact that the PMT plugin is able to take multiple –add-pid Options. This makes coding a lot harder and basically there is no Need for it because as you say, the strenght of tsduck is to add multiple filters or even to pipe. 
Is there a technical reason why the PMT plugin can take multiple –add-pid Options instead of only one?

Greets from Vienna!
Harald

Von: Thierry Lelegard
Gesendet: Freitag, 9. März 2018 23:51
An: tsduck/tsduck
Cc: haraldjordan78; Author
Betreff: Re: [tsduck/tsduck] tsplugin_pmt added --add-pid-id--add-programinfo-id (#40)

Hi Harald,
Thank you for your contribution. I merged it into the master branch. I added your name in the list of contributors (if you don't mind).
I have made a change in your option. I must admit I did not like much the --add-pid-id option because, as you pointed out yourself in your help text, this restricted the number of --add-pid to one. Additionally, it was designed to work only on new PID's and could not set a component tag or cue type on an existing PID.
So, I kept the whole feature but I changed the option. I replaced --add-pid-id with two options --set-stream-identifier pid/id and --set-cue-type pid/type. Each option apply to a specified PID, whether it is a new one or an existing one. We may continue to use many --add-pid and also use many --set-stream-identifier and many --set-cue-type.
I also took that opportunity to do some internal refactoring in the plugin. I wrote it about 10 years ago and did not modify it for years, so it was time for a lifting. Specifically, I moved the processing of the two new options in a template method so that we may add several similar types of options without writing code, just calling the template method.
Thanks again for your contribution and have a nice week-end.
Thierry

haraldjordan78:
What do you think about the PMT plugin just taking a single --add-pid Option so the user has to add multiple -P options or use piping in order to get multiple pids added to the PMT? 
We might need to add an option to disable PMT version patching in this case tough.

lelegard:
Hi Harald,

Principles are good as long as they are correctly balanced. Allowing more than one `--add-pid` option does not change the semantics of the plugin, which is "work on a PMT", since one PMT has more than one component. The syntax of the command is not more complicated, just allow more than one occurence of the option. The extra coding effort is very small. On the other hand, having to pipe 4 plugins to add 4 PID's in one single PMT's adds extra demux & packetization overhead and - more annoying - multiply by 4 the latency of this table, which may have a negative impact on clear <=> scrambled transitions where the CA systems have some constraints in the delays between PMT transitions and ECM's.

What would have been too much would have been to allow to modify multiple PMT's (from different PID's) in the same plugin. Here, the syntax of the command would have been much more complicated to understand. And the coding would be complex and error-prone. And this complexity is not necessary since the demux / repacketization latencies are not added, they are parellelized because the two plugin instances apply to distinct PID's.

This is the balance: work on one single PMT but allow multiple operations on this PMT.

About the template, this one is very basic. If you want to experience the frightening "art of templating", have a look at file `tsArgMix.h`. The class `ArgMix` is a proxy class that you never explicitly manipulate but which allows you to create fully-typed variable-length lists of arguments for `Format` and `scan` in class `UString`. I cannot decide whether I should be proud or ashamed of the templates here. I took me a full day to write this class. You may call it over-engineering but I desperately wanted to create heterogeneous but fully typed variable lists of arguments. I hate the C-style untyped and unsafe `printf` and `scanf`. So many bugs and crashes were caused by unchecked typing. The good side of TSDuck coding is its robustness. I have never encountered any crash, memory leak or multi-thread deadlock with TSDuck. This is only due to the coding of the appropriate base classes or design patterns such as `ArgMix`, `SafePtr` and all forms of _guard_ classes. I am no better than any human programmer, I produce bugs. But, with the help of this tools, most of my bugs are spotted by the compiler or automatically fixed at run time by those base classes.

Waiting for your next pull request...

PR #45, opened 2018-03-23
small bugfix to pmt plugin

haraldjordan78:


haraldjordan78:
sorry i should have added some words for this change. Without this change, the option does not work, because everywhere else the option is called --set-stream-identifier instead of --add-stream-identifier

lelegard:
Hi Harald,

This particular fix in plugin `pmt` is unfortunately incorrect. There are two distinct options:
- An old one, `--add-stream-identifier` which is a boolean one, without parameter. It adds a component tag on all elementary streams which do not have one. The values are allocated sequentially.
- The new one that you contributed to create, `--set-stream-identifier`, which sets a precise component tag value on a precise component.

There are some non-regression tests for these options in `test-020` in the `tsduck-test` repository.

I agree that this is a big confusing. But this is the kind of thing that happens when two modifications are made at 10 years of distance...

So, I have to mark this one as invalid.

I see that you just pushed another commit in the same  PR for the `mux` plugin which seems interesting. So, I would prefer it to be in a different pull request. Especially because one is invalid but not the other one.

If you are not familiar with the PR workflow (and I wasn't either not so long ago), I suggest that you read [this article](https://gist.github.com/Chaser324/ce0505fbed06b947d962). I found it quite useful for me.

Issue #1, opened 2017-09-01
TBS DVB tuners not detected


rezna11:
TBS5925, TBS5922 nor DVBSky S960 are detected as input DVB devices. What DVB tuners are supported?

I have other tuners too, should I try those? Is there a list of compatible tuners?
```

C:\Users\Administrator>tsp -I dvb -d "TBS 5922 USB DVB-S2 Tuner"
* Error: dvb: DVB device "TBS 5922 USB DVB-S2 Tuner" not found

C:\Users\Administrator>tslsdvb -ve
tslsdvb: no DVB device found

=== Device category KSCATEGORY_BDA_NETWORK_PROVIDER

=== Device category KSCATEGORY_BDA_TRANSPORT_INFORMATION

=== Device category KSCATEGORY_CAPTURE

=== Device category KSCATEGORY_TVTUNER

=== Device category KSCATEGORY_BDA_RECEIVER_COMPONENT

device "TBS 5922 USB DVB-S2 Capture"
  Some supported interfaces:
    interface IAMDeviceRemoval
    interface IAMFilterMiscFlags
    interface IAMovieSetup
    interface IBaseFilter
    interface IKsControl
    interface IKsPropertySet
    interface IKsTopologyInfo
    interface IMediaFilter
    interface IMediaSeeking
  IKsPropertySet properties support:
  IKsControl properties support:
  IKsTopologyInfo:
    Categories: KSCATEGORY_BDA_RECEIVER_COMPONENT
    No node found

  - Pin "MPEG2 Transport", direction: input
    Some supported interfaces:
      interface IKsControl
      interface IKsPin
      interface IKsPropertySet
      interface IMemInputPin
      interface IPin
      interface IQualityControl
    IKsPropertySet properties support:
    IKsControl properties support:

  - Pin "MPEG2 Transport", direction: output
    Some supported interfaces:
      interface IAMBufferNegotiation
      interface IAMStreamConfig
      interface IAMStreamControl
      interface IKsControl
      interface IKsPin
      interface IKsPropertySet
      interface IMediaSeeking
      interface IPin
      interface IQualityControl
    IKsPropertySet properties support:
    IKsControl properties support:

device "TBS 5925 DVBS2 Capture"
  Some supported interfaces:
    interface IAMDeviceRemoval
    interface IAMFilterMiscFlags
    interface IAMovieSetup
    interface IBaseFilter
    interface IKsControl
    interface IKsPropertySet
    interface IKsTopologyInfo
[...]

PR #46, opened 2018-03-23
additional options for mux plugin

haraldjordan78:
Dear Thierry,

sorry again for the extra-effort. I did not wish to commit 2 changes in one pull request of course but it seems thats the way how github works. Still learning though.

The following changes to the mux plugin introduce multiple additional options that let the user specify a PCR time where to start and end insertion, as well as a repetition time, e.g. insert a packet only every x milliseconds. 

Again of course in very basic coding style, i hope you still like the changes and do not have to invest too much effor to make it fulfill your standards. 

The new code is of course tested and also i tested some old functions (sure not all), and to work as designed.

The most problematic spot in my humble opinion is line 355, where i enclose your original "read from file into packet" code in my if statement. 
Second problematic is the use of a signal _minmaxpts_rangeOK which makes reading and writing additional functions pretty hard. But unfortunately i could not imagine a much better way besides probably creating a whole new plugin.
Third problematic is the new option --minmaxptspid: it's name might be wrong as  i am expecting a pcr pid actually. Probably this should be named pcrpid and additionally the code should check if this option is needed and error in case it is not defined.

Please let me know your thoughts and i am happy to re-engineer.

Greetings from Vienna!

### DIFF
--- a/src/tsplugins/tsplugin_mux.cpp
+++ b/src/tsplugins/tsplugin_mux.cpp
@@ -66,6 +66,16 @@ namespace ts {
         PacketCounter _inter_pkt;          // # TS packets between 2 new PID packets
         PacketCounter _pid_next_pkt;       // Next time to insert a packet
         PacketCounter _packet_count;       // TS packet counter
+        uint64_t      _inter_time;              // Milliseconds between 2 new packets, internally calculated to PTS (multiplicated by 90)
+        uint64_t      _minpts;                  // Start only inserting packages when this PTS has been passed
+        PID           _minptsPID;               // defines the PID of minpts setting
+        uint64_t      _maxpts;                  // After this PTS has been seen, stop inserting
+        bool          _minmaxpts_rangeOK;       // signal indicates if we shall insert 
+        uint64_t      _max_insert_count;        // from userinput, maximum packages to insert
+        uint64_t      _inserted_packet_count;   // counts inserted packets
+        uint64_t      _youngest_pts;            // stores last pcr value seen (calculated from PCR to PTS value by dividing by 300)
+        uint64_t      _pts_last_inserted;       // stores nearest pts (actually pcr/300) of last packet insertion
+        uint8_t       _last_inserted_cc;        // Continuity counter of last inserted packet
 
         // Inaccessible operations
         MuxPlugin() = delete;
@@ -95,7 +105,12 @@ ts::MuxPlugin::MuxPlugin(TSP* tsp_) :
     _bitrate(0),
     _inter_pkt(0),
     _pid_next_pkt(0),
-    _packet_count(0)
+    _packet_count(0),
+    _inter_time(0),
+    _minpts(0),
+    _minptsPID(0),
+    _maxpts(0),
+    _inserted_packet_count(0)
 {
     option(u"",                       0,  STRING, 1, 1);
     option(u"bitrate",               'b', UINT32);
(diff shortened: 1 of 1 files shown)


lelegard:
Hi Harald,

Thank you for your pull request. It looks really nice and I will review and integrate it. Probably not today though.

Concerning the 2 fixes in 1 PR, do not worry. The PR workflow (and even git in general) is not straightforward at the beginning. One thing to remember, as highlighted in the [article](https://gist.github.com/Chaser324/ce0505fbed06b947d962) I previously mentioned, is to create a distinct branch per new feature (future PR), in your forked repository. Thus, the 2 features can be pushed and merged independently.

Cheers from Paris !

lelegard:
And injecting data based on PTS values is a perfume of SCTE 35 ... :)

haraldjordan78:
Hehe "a perfurme of SCTE 35" is well expressed. Possibly it is worth to introduce a new plugin for SCTE insertion but as the mux plugin does not have too many options, my additions would fit into it. Whenever you have the time to review, please just let me know any doubt.

lelegard:
Hi Harald,

I am currently integrating your code. I understand that you use PTS = PCR / 300. But, formally, splicing cues are based on PTS values from the PES headers, not PCR in TS AF. In practice, it is true that PTS and PCR's are fully synchronised, but not always in the same PID. Sometimes, the PCR PID of a service is on its own, carrying nothing else than PCR's. So, in case the user specifies by mistake the video PID of the service instead of the PCR PID, if the two PID's are different, this won't work. I will make the following change: when collecting time stamps from packets, I take the one I find, PCR (/300) or PTS, whichever is present. What do you think?

lelegard:
Hi Harald,

I have merged your PR. I have made a few modifications:

- Renamed options for consistency with others: `--inter-time`, `--max-insert-count`, `--min-pts`, `--max-pts` and `--pts-pid`
- I fixed a logic error near the end. It would not work when the input file contains packets from distinct PID's.
- I removed your processing of the CC, this was already done.

Could you please validate if the modified version passes your tests ? I will wait for your answer before releasing version 3.10, just in case I introduced a bug.

Thank you Harald for your work.

haraldjordan78:
Hi Thierry,

you are too kind. Thanks a lot for all the time you invest to this. I checked in detail the differences between my original modifications and your enhancement. Sure everything looks ok for me. 

It is a smart idea of you to just default to the last PCR in case the --pts-pid option is 0. Sure this only works for SPTS but i guess most tsduck users deal with SPTS instead of MPTS and the ones who do have a MPTS need to be aware of their PCR PID anyway. 

Please give me a week to confirm the new version is working as designed but from the code i could not see any doubt about that. 

Spending about an hour to diff your changes (wanted to be sure i got everything) was cosmetical enhancements:
move line 351 to  307
move line 385 to 366 

The thing i liked most was your change to my #1 problematic spot i mentioned originally. I was aware that this is a problematic spot but i could not imagine to do negative tests instead of positive tests. But in the end your check makes everything more clear and readable, even if negative tests are generally teached as taboo.

OLD:
if (_minmaxpts_rangeOK && (_max_insert_count > _inserted_packet_count || _max_insert_count == 0)) {

NEW: 
if (!_pts_range_ok || (_max_insert_count != 0 && _inserted_packet_count >= _max_insert_count)) {

I regret that you did have to invest that much time for my changes and feel pleased that you did. It teaches me great lessons!
Please give me a week to confirm the new code works in production. If you like, i can update the word documentation as soon as everything is released.

All the Best,
Harald

lelegard:
>  i guess most tsduck users deal with SPTS instead of MPTS 

Certainly not. I have used TSDuck for years mostly on MPTS. Each time you get a stream from a DVB tuner, satellite, terrestrial, etc, this is an MPTS. I hardly used SPTS outside IP-TV. So, being prepared for all kinds of TS is essential.

PR #65, opened 2018-06-08
tsecmg: max-comp-time parameter was not used

maxha651:


maxha651:
No problem, thanks for a really nice piece of software :+1:

PR #102, opened 2018-07-19
timeref: Add support for --start "system"

lars18th:
For the plugin `timeref` the parameter `--start` has now a new predefined value `"system"`. This value will read the system clock and uses it.


lelegard:
Hi @lars18th 

Thank you for your last two contributions. I maintain a list of people who contributed to the project in a file named CONTRIBUTORS.txt. If you would like to add your name here (and you deserve it), please send me your name at contact@tsduck.io. Note that only real names are added.

PR #111, opened 2018-08-22
Add decimal view of pids in service analysis

chibbert:
Most places in the analysis output the pids are in hex with decimal value in brackets.

This is missing in the service analysis view.

I couldn't get everything to fit in the default 79 character width, so I've currently added this as an option "--service-analysis-decimal-pids" which you add in addition to "--service-analysis" and it uses a wider grid.

If you think it is acceptable to just increase the grid size for all cases from 79 to 94 then we can get rid of the parameterisation of this and will very much simplify the code changes.

Example output:

image


lelegard:
Hi @chibbert 

Thank you for your contribution. I am integrating it.

I prefer to keep the default to 79 columns since many terminal emulators have a default width of 80. Moreover, depending on the terminal configuration, displaying an actual 80th character may force a new line, creating a visually ugly empty line. All these defaults were here only to perfectly emulate physical terminals such as VT-100 series from the 80's (or even 70's maybe) :cry: 

This was a dilemma when I originally formatted the output. What could fit in 79 columns? You are right to add a wider display with more information.

To be more general, I renamed the option into `--wide-display` so that we may add more information on other lines later. Maybe.

lelegard:
Hi @chibbert 

One recommendation: To be able to submit PR's, please do not use the branch named `master` to develop new features. By doing commits here, you diverge from the main repo, using merge and merge again. Please keep `master` aligned with the main repo, create temporary branches named from the feature (e.g. `decimal-pids`), develop your modifications there and create the PR's from these temporary branches. There is a good gist [here](https://gist.github.com/Chaser324/ce0505fbed06b947d962) explaining this.

lelegard:
Merged. Not directly seen as merged on GitHub because of subsequent conflicts in branch `master` in forked repo.

PR #146, opened 2018-10-23
DektecInputPlugin:  fix bug and add feature; Zap plugin:  add feature

alevinsn:
This fixes a bug with the Dektec input plugin that prevented it from being used.  The timeout-related option is named "receive-timeout", yet it was expecting an option named "timeout".  Now, it expects an option named "receive-timeout".

Signed-off-by: Aaron Levinson <alevinsn_dev@levland.net>


alevinsn:
Note that I just added a new, totally unrelated commit to my fork, to the zap plugin.

PR #152, opened 2018-10-31
Various small changes

alevinsn:
Might be good to carefully review implementation of SplicingPointOffset() based on MPEG-TS specification, but I think I got it right.


lelegard:
Hi Aaron,

Thanks for your contribution.

I had to modify a few things:
- One of the size checks was incorrect (off by 1) in getting the splice countdown offset in TSPacket.
- In plugin `pes`, the SEI type values were in range 0-31 (and in a 32-bit bitset). In fact, there is no defined size for SEI type (strange AVC variable-length integers) and predefined values exceed 31. Now using 32-bit values (a reasonable size).

Concerning your comment about TSPacket and caching offset values in member variables, we can't because `TSPacket` is a "fake" class. Its implementation must be restricted to the 188-bytes data, no other member, no vtable, etc. This is required to map raw packet buffers to vectors of `TSPacket`.

alevinsn:
Regarding the size check, I reviewed your changes, and they look correct to me.  The other changes look good to me as well.  Looking at the code, technically, I think the .empty() check is unnecessary, although perhaps slightly more efficient in the case that the set is empty.

Regarding TSPacket--that makes sense, although you could have another class that is used to wrap a TSPacket that does this kind of thing.

PR #156, opened 2018-11-07
Added ability to use %f in format strings and made use of it in rmsplice plug-in

alevinsn:
I thought it wouldn't work this way, since there is already a pull request for various rmsplice changes to add support for immediate splice events, but this pull request is bringing in those changes as well, since they haven't been committed to master yet.  I needed to base these changes on the rmsplice immediate changes, since I wanted to change some log messages in my version of the rmsplice plug-in.


lelegard:
Hi Aaron,

Thank you for your contribution. I appreciate the quality of coding (especially in minefields such as `ArgMix` and `UString::format`).  I am integrating it. I have just a few details to fix (warnings with LLVM on macOS) and the doc to update.

alevinsn:
Thanks--all the changes that you've made to improve this set if commits look good.  One thing that might be desirable is to add support for doubles to ArgMixOut and %f to scan, for consistency.

Issue #166, opened 2018-11-15
BUG: plugin ENCAP initial start inconsistencies


lars18th:
Hi @lelegard ,

After a lot of testing (several days), I found a very strange bug in the `encap` plugin.

The trouble is with the starting packets. In particular with the **first** packet. Without using the `--pack` mode, without PCRs, and with sufficient NULL packets, the expected beheviour is:

- One packet has the PUSI flag set, the byte 5 is `0` as it points to an starting packet, and the payload has 184 bytes of the original packet.
- And the following packet doesn't have the PUSI flag, and the Adaptation padding stuff occupies 180 bytes, so it only leaves 4 bytes at end of the original packet.

This is right. And it works with _all_ packets... **except** with the _first_ and _second_ packet. Instead of it:

- The first packet, instead of `0` in byte [5] it has a `4`. And the payload (the start of the packet) doesn't starts in the position {4} (byte [6] of the TS packet) but in the byte [12] (0x0C in the TS packet). And the bytes between [5]-[11] have random data.
- And the second packet, instead of 4 bytes at end of payload, it has 8 bytes.

I checked the code for an initialization error, but I can't find anything.
Any ideas?
Please, help me!

lars18th:
Hi @lelegard ,

This is a **very obscure bug**!!! :open_mouth: 
In fact, using _only_ the encap filter, the problem doesn't appears.
Everything seems to point to a memory leak inside the library.

Fortunately I have found a test that reproduces the error.
It uses the "test-001.ts" file from the tesduck-test:
https://github.com/tsduck/tsduck-test/blob/master/input/test-001.ts

Here a command that **doesn't** fail:
```
tsp  --add-input-stuffing 1/20 \
 -I file test-001.ts \
 -P duplicate 0-18=8000 \
 -P encap -p 8000-8018 -o 8032 \
 -P svrename -p "EDIT" 8801 \
  -O file out-no-error.ts
```

And this command produces these two first packets of pid 8032 (the one of the capsulation):
```
Packet: 2603
File Offset: 489364
Time: 00h00m00:078
?

Header:

sync_byte: 0x47 (71)
transport_error_indicator: 0
payload_unit_start_indicator: 1
transport_priority: 0
PID: 0x1F60 (8032)
transport_scrambling_control: 0 (not scrambled)
adaptation_field_control: 1 (no adaptation_field, payload only)
continuity_counter: 0x6 (6)

Data:
[...]

PR #167, opened 2018-11-16
Fixes ugly optimization

lars18th:
When packet processing starts, in this moment the value of _lateIndex is 0. And when the first packet is queued, this value changes to 1. However, some optmization breaks this, and only after the flush of the queue the value returns to 1.

So one simple solution to fix this trouble is to set it to 1 at starting time. In any case, when the loop empties the queue, the value is set to 1 (see last line of function fillPacket() ).

This resolves the ugly output of initial packets.

### DIFF
--- a/src/libtsduck/tsPacketEncapsulation.cpp
+++ b/src/libtsduck/tsPacketEncapsulation.cpp
@@ -169,6 +169,7 @@ bool ts::PacketEncapsulation::processPacket(TSPacket& pkt)
     PID pid = pkt.getPID();
     bool status = true;  // final return value.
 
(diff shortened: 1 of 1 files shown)


lelegard:
Closed as not applicable. See discussion in #166.

lars18th:
> Closed as not applicable. See discussion in #166.

Please, read https://github.com/tsduck/tsduck/issues/166#issuecomment-439599019 and reopen it.

lelegard:
I have read #166 and explained there why I will not reopen it. In fact, the patch does nothing.

Issue #157, opened 2018-11-07
BUG: plugin ENCAP doubles bitrate


lars18th:
Hi @lelegard ,

Last days I initiated a deep testing of the `encap`/`decap` plugins. And I found a bug:

- The `encap` plugin doesn't does the task described in the documentation, and commented in #126 .
- At time, the plugin is _doubling_ the number of input packets. In fact, it generates all the time **two** packets for each input packet. The result is one packet with payload only, with the next packet with the adaptation field set with 180 bytes size (and mainly empty with zeroes) with the remaining bytes of the original packet as payload.
- So you see: Input={A...AAA....AA....} encapsulated to Output={A'A''...A'A''A'A''A'A''....A'A''A'A''....}.

Please, can you check the algoritm?

lars18th:
Hi @lelegard ,

Your explanation is crystal clear!
Thank you for your support. I'll test it.

Regards.

lars18th:
> Please, pull, rebuild and tell us if option --pack fits your needs.

Hi @lelegard ,

I see some strange results using the `--pack` option!

First of all, I need to say that I'm using (at time) the `encap` filter to pack PSI/SI tables. So my current command is `tsp ... -P duplicate 0-21=600 -P encap -p 600-621 -o 622 --pack ...`

So, adding the option `--pack` nearly all output packets have PUSI set. I feel this a bug, as I understand that only the _first_ packet of a "pack" needs to have the start indicator set to 1; while the rest doesn't have it. Right?

Futhermore, it's strange that the **first** output (outher) packet of the pack, instead of having as payload the content of **one** inner packet it has parts of **two** inner packets. My understanding is that the first output will have the first inner packet, except the last 4 bytes that have no space in the packet. Right?

Besides these problems (please review the code), I have a suggestion: Your first implementation without the `--pack` option seems to be the smartest strategy. However, it fails when the original stream is mulplitexed. In this case, it's common to have "holes" between packets with the same pid. Then, if some NULLs are in between of two sequential packets of the same pid, but they aren't adjancent, then an additional packet is allocated. That's the case where the `--pack` option has sense. However, I feel it's good not to put together packets that are far away.

So I suggest to add the concept of _distance_ in between for the parameter `--pack`. As example: `--pack 64` will indicate a _maximum_ number of 64 foreign consecutive packets before force to close the "pack" and flush the remaining buffer. See this example:

`.....A...A.A.............................A.A..A..A..AAAA.......`

[...]

PR #171, opened 2018-11-19
PES encapsulation implementation

lars18th:
Hi,

I present here a _second_ alternative format for the encapsulation (`encap` and `decap` filters).

You can found a deep description in the comment included in the header. See it here:
https://github.com/lars18th/tsduck/blob/encap-pes-mode/src/libtsduck/tsPacketEncapsulation.h#L68

You can now use for the Encapsulation the new `--pes-mode 1` (recommended) and `--pes-mode 2`; both w/o `--pack`. The Decapsulation is entirely automatic.

I created this mode because the FFmpeg tool can't use the plain encapsulation. It only process (or copies) PES packets. Therefore, this new encapsulation is necessary when used in combination with FFmpeg.

For sure, this encapsulation is **fully** tested:
1. It pass the test https://github.com/tsduck/tsduck-test/blob/master/tests/test-035.sh
2. Multiple `encap`+ `decap` files are tested with `--pes-mode 1`, `--pes-mode 1 --pack`, `--pes-mode 2` and `--pes-mode 2 --pack`.
3. Previous files are processed with FFmpeg in different ways, and after the `decap` filter is used to recovery the original content. Afterwards it's checked using the tool `tscmp`.

All problems are solved! The code is robust and tested.
Regards.

### DIFF
--- a/src/libtsduck/tsPacketDecapsulation.cpp
+++ b/src/libtsduck/tsPacketDecapsulation.cpp
@@ -94,10 +94,114 @@ bool ts::PacketDecapsulation::processPacket(ts::TSPacket& pkt)
     // Where to look at in input packet. Start at beginning of payload.
     size_t pktIndex = pkt.getHeaderSize();
 
-    // Get pointer field in PUSI packets.
-    const bool pusi = pkt.getPUSI();
-    const size_t pointerField = pusi && pktIndex < PKT_SIZE ? pkt.b[pktIndex++] : 0;
-    if (pusi && pktIndex + pointerField > PKT_SIZE) {
+    // when PLAIN encapsulation is used corresponds to PUSI;
+    // and when using the PES encapsulation it's an internal flag.
+    bool start_mark = false;
+
+    // A special case may arise when one original PES packet is fragmented
+    // and the pointer to the next inernal packet points to a position in the
+    // second part of the packet. This offset solves the problem.
+    // However, it's good to overcome the fragmentation!
+    uint8_t pes_fragment = 0;
+
+    // Differentiate whether it's a plain encapsulation or a PES encapsulation
+    if (!pkt.getTEI() && pkt.isClear() && pkt.hasPayload()) { // General checks
+        if (pkt.getPUSI() && pktIndex < (PKT_SIZE - 9) &&
+            pkt.b[pktIndex]   == 0x00 &&
+            pkt.b[pktIndex+1] == 0x00 &&
+            pkt.b[pktIndex+2] == 0x01) {
+            // PES header found, continue...
+            pktIndex += 3;
+
+            // Check for correct Type Signature
+            if (pkt.b[pktIndex++] != 0xBD) {
+                return lostSync(pkt, u"invalid PES packet, type differs");
+            }
+            // Private Stream 1 found, continue...
+
+            // Check for PES Size and valid Flags
+            if (pkt.b[pktIndex++] != 0x00) {
+                return lostSync(pkt, u"invalid PES packet, size incompatible");
+            }
+            const size_t pes_size = pkt.b[pktIndex++];
+            // 178 bytes is the maximum PES packet size in origin.
+            // However, if an external processor splits the packet and inserts
+            // some PES header data (like PTS marks), then the size increases.
+            // We see PES lengths of 189, but a more conservative value is used.
+            if (pes_size > 255 || pes_size < 18) {
+                return lostSync(pkt, u"invalid PES packet, wrong size");
+            }
+            if (pkt.b[pktIndex] != 0x80 && pkt.b[pktIndex] != 0x84) {
+                return lostSync(pkt, u"invalid PES packet, incorrect flags");
+            }
+            pktIndex++; // Ignore these flags
+            pktIndex++; // Ignore these flags
+            size_t readed_size = pkt.b[pktIndex++]; // PES header size
+            if (readed_size > 0) {
+                // Advance up to the end of the PES header
+                pktIndex += readed_size;
+                pes_fragment = readed_size; // When fragmentation appears in the outer packet, this offset is added to checks.
+            }
+            // PES header OK!
+
+            // Start reading KLVA data...
+            if (pktIndex > PKT_SIZE - 18) {
+                return lostSync(pkt, u"invalid PES packet, data unknown");
+            }
+            // Check for our KLV correct KEY
+            // UL Used: 060E2B34.01010101.0F010800.0F0F0F0F
+            // This is an Unique ID in the testing range.
+            if (pkt.b[pktIndex]    != 0x06 && pkt.b[pktIndex+1]  != 0x0E && pkt.b[pktIndex+2]  != 0x2B && pkt.b[pktIndex+3]  != 0x34 &&
+                pkt.b[pktIndex+4]  != 0x01 && pkt.b[pktIndex+5]  != 0x01 && pkt.b[pktIndex+6]  != 0x01 && pkt.b[pktIndex+7]  != 0x01 &&
+                pkt.b[pktIndex+8]  != 0x0F && pkt.b[pktIndex+9]  != 0x01 && pkt.b[pktIndex+10] != 0x08 && pkt.b[pktIndex+11] != 0x00 &&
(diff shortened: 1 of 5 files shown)


lars18th:
Hi @lelegard ,

I hope this PR will be merged.
I have tried to comment as much as possible on the code to make it understandable. Of course, you could have written the code in a more compact way. But I think it's enough as it is.

Now, only remains to add the option `--pack [number]` (aka _distance_) commented here https://github.com/tsduck/tsduck/issues/157#issuecomment-437369120

Best.

lars18th:
Hi @lelegard ,

> Now, only remains to add the option...

Ready to merge!
The "distance" is now included: 0=unlimited; 1=like no pack; 2 or more is the distance (ex. a distance of 2 equals to 1 packet in between).

Please, test it. :wink:

lelegard:
Hi @lars18th 

Thank you. I will review the code when I have more time that right now.

I have read the article about KLV. It seems a more complicated form of TLV. I understand the need for PES packets because of ffmpeg. But do we need compatibility with this standard? This seems a lot of complexity for an unclear benefit.

I have read the comments (but did not try to understand the code yet). The comments give details but the broad picture is unclear.

- How many inner TS packets do you encapsulate in one PES packet?
- I do not understand the sentence _"Each TS packet contains only one key, with a size no larger than the payload of one TS packet. So each PES packet fits into a single TS packet."_ This would mean that the PES packet is too short to encapsulate an inner TS packet. Could you explain ?

Concerning the distance, it is good to have added the feature. However, it is not so good to have broken the compatibility of option `--pack` since it is now refused as a flag alone (without value). I would have made the value optional, allowing usages such as `--pack` or `--pack=100`.

lars18th:
Hi @lelegard ,

> I have read the article about KLV. It seems a more complicated form of TLV. I understand the need for PES packets because of ffmpeg. But do we need compatibility with this standard? This seems a lot of complexity for an unclear benefit.

It's not true at all, the KLV isn't very complex. It's a "simple" PES encapsulation of _ANY_ kind of metadata. The reason for using it is because it's the _only_ metadata supported in the FFmpeg tool. In fact, all drone metadata uses the KLV format. And FFmpeg has a lot support for that format. There is no other reason.

> How many inner TS packets do you encapsulate in one PES packet?

The PES encapsulation is just an envelope. Futhermore, the envelope is always on the top. So any TS packet has the form: TS Header + PES Header + Payload. And the _payload_ is the data as with regular encapsulation. The only particularity is that we generate one PES packet for each TS packet. This requirement is because we need to write the PES length in the header. However, after that, any external tool can process the packet and fragment it. But, this doesn't break the decapsulation process. It works as expected. And as for your question: we encapsulate one inner packet in two PES packets.

> I do not understand the sentence "Each TS packet contains only one key, with a size no larger than the payload of one TS packet. So each PES packet fits into a single TS packet." This would mean that the PES packet is too short to encapsulate an inner TS packet. Could you explain ?

:smile_cat:  A picture better than a thousand words:  --pes-mode 1 (no pack)
```
         0001 0203 0405 0607 0809 0A0B 0C0D 0E0F 0123456789ABCDEF
0x000000 475F 6039 1E00 FFFF FFFF FFFF FFFF FFFF G_`9............
0x000010 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000020 FFFF FF00 0001 BD00 9384 0000 060E 2B34 ..............+4
0x000030 0101 0101 0F01 0800 0F0F 0F1F 7F00 5F00 .............._.
0x000040 1D00 00B0 2500 10CF 0000 0000 E010 00BA ....%...........
0x000050 E44C 00BB E4B0 00BC E514 00BD E578 00BE .L...........x..
0x000060 E5DC 00BF E6A4 84ED 0368 FFFF FFFF FFFF .........h......
0x000070 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000080 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000090 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x0000a0 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x0000b0 FFFF FFFF FFFF FFFF FFFF FFFF           ............

         0001 0203 0405 0607 0809 0A0B 0C0D 0E0F 0123456789ABCDEF
0x000000 475F 603A 6000 FFFF FFFF FFFF FFFF FFFF G_`:`...........
0x000010 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000020 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000030 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000040 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000050 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000060 FFFF FFFF FF00 0001 BD00 5184 0000 060E ..........Q.....
0x000070 2B34 0101 0101 0F01 0800 0F0F 0F0F 3DFF +4............=.
0x000080 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x000090 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x0000a0 FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF ................
0x0000b0 FFFF FFFF FFFF FFFF FFFF FFFF           ............
```
Two consecutive outer packets with just **one** inner packet. Note that two PES packets are used.

> Concerning the distance, it is good to have added the feature.

Thank you! :wink: 

> However, it is not so good to have broken the compatibility of option --pack since it is now refused as a flag alone (without value). I would have made the value optional, allowing usages such as --pack or --pack=100.

Please, can you do it? I'm sure it will be trivial for you. But I don't see any examples of such parameters in your code.

Regards.

lars18th:
Hi @lelegard ,

Another comment: Using PES mode 1 without pack, the total overhead is identical to the current (plain) encapsulation without pack. Therefore, even if the header is larger, the result is the same. Only when the "pack" mode is activated is the overhead noticeable.

lelegard:
Hi @lars18th 

In the last commit, I have added low-level methods in class `TSPacket` such as `init`, `createPCR`, `setPayloadSize`. They should help. This avoid messing with bits and bytes in packet headers and only rely on packet methods. I already modified `PacketEncapsulation` (committed version, plain only) to use them.

lars18th:
Hi @lelegard ,

I'm sorry about the delay! I've been sick for two weeks. :cry: 

> I am trying to merge the PR but there are too many problems for now:

I'll invetigate it! For me this enhancement is a must-have. And I spent a lot of effort to achieve it. So I really need to fix it and then you can merge it with the mainstream.

> Crash when using --pes-mode with --pcr-pid. In fact, this is an assertion failure. The build of the combination of header / adaptation field / PES header is not robust enough and fails when a PCR is present.

Just the only case (with pcr) that I haven't _exhaustively_ checked. :wink: Sorry!

> Incorrect check of the KLV key in decapsulation (inversions of && and ||).

Oh là là! :fearful: 
Sorry! Literally a beginner's mistake.
Thank you so much for noticing it. :+1: 

> The code does not compile on Windows because of invalid mixing of type sizes (use size_t when something is a size, not uint8_t, then explicitly convert to uint8_t when setting a header byte).

I'll try to compile it also in Windows (for the development task I prefer Linux... more fast and simple: just a `make` call. I hate the MSVC!).
Sorry for this error!

> In the last commit, I have added low-level methods in class TSPacket such as init, createPCR, setPayloadSize. They should help. This avoid messing with bits and bytes in packet headers and only rely on packet methods. I already modified PacketEncapsulation (committed version, plain only) to use them.

I'll try to rebase and use these new methods.
Be on hold!
Regards.

lelegard:
> I'll try to compile it also in Windows (for the development task I prefer Linux... more fast and simple: just a make call. I hate the MSVC!).

I also prefer macOS or Linux as a development platform for the same reasons and share your distaste for MSVC. However, GCC/LLVM and MSVC are distinct worlds and, even for portable code, they produce distinct warnings and spot distinct potential bugs. This is why I always test on Windows at the end. GCC and LLVM usually share the same characteristics, although they occasionally generate distinct warnings.

Waiting for your update.

lars18th:
Hi @lelegard ,

Sorry for the delay. A new version is provided.
Please, review it https://github.com/tsduck/tsduck/pull/182 .

PR #172, opened 2018-11-20
tsPlatform: add support for aarch64

athoik:
This commit adds support for aarch64 (aka ARM64).
The memory barier copied from here: https://github.com/mirage/shared-memory-ring/blob/master/lib/barrier.h#L32

### DIFF
--- a/src/libtsduck/tsPlatform.h
+++ b/src/libtsduck/tsPlatform.h
@@ -296,6 +296,10 @@
     //!
     #define TS_ARM
     //!
+    //! Defined when the target processor architecture is ARM64.
+    //!
+    #define TS_ARM64
+    //!
     //! Defined when the target processor architecture is STxP70.
     //!
     #define TS_STXP70
@@ -348,6 +352,13 @@
     #if !defined(TS_ADDRESS_BITS)
         // To be fixed for ARM 64
         #define TS_ADDRESS_BITS 32
+    #endif
+#elif defined(__aarch64__)
+    #if !defined(TS_ARM64)
+        #define TS_ARM64 1
+    #endif
+    #if !defined(TS_ADDRESS_BITS)
+        #define TS_ADDRESS_BITS 64
     #endif
 #elif defined(__stxp70__) || defined(__STxP70__)
     #if !defined(TS_STXP70)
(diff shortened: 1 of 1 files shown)


athoik:
With above changes I was able to compile tsduck using the following recipe for [openembedded](http://www.openembedded.org/wiki/Main_Page).

```
 SUMMARY = "MPEG Transport Stream Toolkit"
DESCRIPTION = "TSDuck is an extensible toolkit for MPEG/DVB transport streams.\
        TSDuck is used in digital television systems for test, monitoring, integration, debug, lab, demo."
MAINTAINER = "https://tsduck.io/"
LICENSE = "BSD-2-Clause"
LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=cd9752bb95f52810fa1e609da531fa48"

SRC_URI = "git://github.com/tsduck/tsduck.git file://ts.patch"
SRCREV = "${AUTOREV}"

DEPENDS = "gettext curl"

S = "${WORKDIR}/git"

inherit gitpkgv autotools-brokensep pkgconfig gettext

PV = "3.16+gitr${SRCPV}"
PKGV = "${GITPKGVTAG}"
PR = "r0"

EXTRA_OEMAKE = "NOTEST=1 NOPCSC=1 NODTAPI=1 MAIN_ARCH=${TUNE_PKGARCH} SYSROOT=${D}"
```

I had to define the MAIN_ARCH else it was trying to use march=x86-64.

Next I will try to compile under MIPS (mips32el) and ARM (armv7a).

Thanks for your great job.

PS. Still didn't figure how I can use tsp -I dvb to analyze tuned data (eg previously I use to do dvbsnoop PID to analyze one pid), but tsanalyze is just great!

lelegard:
Hi @athoik 

Thanks a lot for your contribution. I do not have any ARM64 system. I only have a Raspberry Pi with Raspbian, ie. 32-bit OS on a 64-bit ARM CPU.

> I had to define the MAIN_ARCH else it was trying to use march=x86-64.

What is your output for `uname -m` and `uname -a` ? I will update `Makefile.common` accordingly so that specifying MAIN_ARCH will be no longer necessary.

> Next I will try to compile under MIPS (mips32el) and ARM (armv7a).

32-bit ARM should be ok (tested on my Raspberry Pi). You may expect modifications to be done on MIPS (not tested yet, not by me at least).

> Still didn't figure how I can use tsp -I dvb to analyze tuned data

Do you mean that you did not manage to tune or you do not know how to analyze the TS you tuned to? In the latter case, I routinely use commands like this one:
~~~
tsp -I dvb ... -P until --seconds 60 -P analyze -o analysis.txt -O drop
~~~
This command analyzes the tuned TS during 60 seconds, produces the same report as `tsanalyze` and then exits. The command `tsanalyze` and the plugin `analyze` are identical (same options, same code).

For automated analysis scripts, I recommend the option `--normalized` for the `analyze` plugin. It produces an output format which can be easily manipulated by `sed` and  `grep` (or `awk` or whatever you like). In the past, I implemented the complete monitoring of DVB networks using simple shell scripts.

PR #182, opened 2018-12-20
PES encapsulation implementation [v2]

lars18th:
This is a new version of the PES mode Encapsulation.
Changes:
- Uses the new function `setPayloadSize()`, so the encapsulation function is more simple.
- Fixed crash when using with `--pcr-pid`.
- `size_t` changes to be more compatible.


lelegard:
> Regarding the support of --pack instead of --pack=0 (both are in fact the same), I don't know how to implement parameters with optional values. Please, implement it by yourself.

Just look [here](https://tsduck.io/doxy/classts_1_1_args.html#ae7aab594c5d06350699bf3d1352ca78a). The profile of the function is:
~~~
        Args& option(const UChar* name = nullptr,
                     UChar        short_name = 0,
                     ArgType      type = NONE,
                     size_t       min_occur = 0,
                     size_t       max_occur = 0,
                     int64_t      min_value = 0,
                     int64_t      max_value = 0,
                     bool         optional = false);
~~~
So, simply change
~~~
option(u"pack", 0, INTEGER, 0, 1, 0, UNLIMITED_VALUE);
~~~
into
~~~
option(u"pack", 0, INTEGER, 0, 1, 0, UNLIMITED_VALUE, true);
~~~

I will look at the PR when I have some time.

lars18th:
> > Regarding the support of --pack instead of --pack=0 (both are in fact the same), I don't know how to implement parameters with optional values. Please, implement it by yourself.
> 
> So, simply change...

Done! Thank you for the tip. :wink: 

> I will look at the PR when I have some time.

OK. I hope you agree with it.

PR #185, opened 2018-12-29
Add --remove-existing-from-new option to svrename plug-in

alevinsn:


lelegard:
Hi Aaron,

Thanks for the contribution. If I understand it correctly, you want to solve the issue of `-P svrename 111 --id 222` where you rename service id 111 into 222 and a service the id of which is 222 already existed. In that case, the new option removes the previous service 222 before renaming 111 into 222. Am I right?

In that case, the plugin `svrename` does two distinct things, removing and renaming. This is typically against the philosophy of TSDuck where each plugin shall remain simple and perform one single function when possible. Or course, there are cases where it is mandatory to perform multiple coordinated actions. But, in that specific case, this was already possible using:

~~~
tsp -P svremove 222 --ignore-absent -P svrename 111 --id 222
~~~

So, unless I missed something, the philosophy of TSDuck dictates to [avoid making a plugin more complex when the same operation is already possible using a combination of elementary plugins](https://tsduck.io/doxy/devplugin.html#pdevworkflow).

alevinsn:
OK, makes sense.   I'll close the pull request.

PR #232, opened 2019-02-22
Windows ATSC detection and signal stats/modulation fixes

oviano:
There are two Windows-specific fixes here:

1. ATSC Detection

When checking tuning spaces there was no code to deal with ATSC; it was only checking to see if a space was DVB-S/C/T compatible. So I've added/restructured the code to deal with this. In addition, to get it working it also needed the equivalent CreateLocatorATSC function adding.

2. Signal stats/modulation

I couldn't understand why other applications were able to show signal strength and quality but not tsduck. It turned out the problem was rather simple; the IID_IBDA_Topology GUID definition was apparently incorrect, so the code could never obtain all the nodes and thus couldn't search for the signal stats/modulation info ones.


lelegard:
Thanks a lot, I will look into it asap.

oviano:
There are a couple of knock-on effects that you might want to consider.

1. signalLocked now also works, so you might want to re-enable the commented out code in tsscan that assumes true for Windows.

2. I think the actual range/meaning of signalStrength is not clearly defined and may vary from one hardware device to another. For me I'm seeing raw values of between 30-40. Using the calculation in tsTuner.cpp for Windows, this always converts to 100%. However, my signal quality is between about 55 and 70, which is border line for receiving reception. So I reckon for my card, the raw signal strength values might already be percentages. This will have implications for your tsscan code where it checks for minimum threshold values etc. Maybe you should just disable the use of strength when scanning on Windows, and just use signal quality?

PR #263, opened 2019-04-14
Add EditorConfig support to project

alevinsn:
The use of EditorConfig (https://editorconfig.org/) is generic and not specific to a particular build environment, but I found this to be helpful, particularly in Visual Studio, because I was constantly having to manually change the C++ tabbing options under Tools->Options->Text Editor->C/C++->Tabs as I switched back and forth between projects that use tabs and projects the use spaces.  EditorConfig support is built into Visual Studio 2017, but for Visual Studio 2015, I had to add the EditorConfig VS2015 extension.  There are other EditorConfig properties that can be set that may be useful, although some of these, such as those pertaining to line endings, are already addressed via the .gitattributes file.


lelegard:
Excellent idea. I was not aware of EditorConfig.

Just one question: How does VisualStudio locate the `.editorconfig` file? Here, it is at the root directory of the project on a `git` standpoint. But the VS project file is in the `build/msvc2017` subdirectory. So, for VS, the project directory is here.

lelegard:
I should have searched before asking the question:

> As documented, it traverses parent directories of the project directory until it encounters a .editorconfig that has root=true.

Very clever and useful indeed.

alevinsn:
Yeah--I was looking for an option to manage this the day I submitted the pull request, because I had found it a pain to make sure that the text editor options are configured correctly.  There are a number of ways to accomplish this, but I think the EditorConfig route, a passive option, works the best.  It is actually tied to the currently active file's location, so if I'm working in my own project and step into TSDuck code, if you go to Tools->Options->Text Editor->C/C++->Tabs, you can see it immediately changes the settings there to reflect those used for the currently active file.  So, if I were to accidentally make changes to a TSDuck file from the wrong workspace (something I try to avoid but that I do accidentally from time to time), the changes will still be formatted correctly.

PR #271, opened 2019-04-22
Synchronous PES mode

lars18th:
The PES mode encapsulation supports now the Synchronous mode.

This mode includes PTS marks in the PES packets. However these marks are only useful for external tools. So, don't use this mode except if you need to remux with an external tool that **requires** to have PTS in Metadata packets. In any other case, it's a waste of space.


lars18th:
Hi @lelegard ,

I'm not sure if the plugin "-P pmt --add-programinfo-id 0xAABBCCDD" is incorrect, but I changed the order in the example as the output is only valid when reversing it.

I hope you test it soon and want to merge it!
Regards.

lars18th:
Hi @lelegard ,

Please review this PR. I'm using it for half a month without troubles.
My first PES implementation lacks the SYNC mode, but it's required in some scenarios (for example for FFMPEG).

And, please consider to comment about the `-P pmt --add-programinfo-id 0xAABBCCDD` described in #274. In fact, the current command commeted in the documentation for the descriptor of the PES encapsulation doesn't work! We need to add it to the stream, not the program.

lars18th:
Hi @lelegard ,

Another comment:
- Decapsulation is not affected. It continues to be transparent. And previously encapsulated streams, continue to work. It only add a "new" mode (PES Sync) that's an enhancement of the previous one.

Regards. :wink:

PR #283, opened 2019-05-01
New extended functions in filter plugin

lars18th:
Added parameters:
 --every: The condition is true only after N occurrences.
 --mark-only: instead of filter out, mark the packets when true.
 --only-with-mark: check only marked packets.
 --in-place-process-remap: instead of filter out, remap the pakets.


lars18th:
Hi @lelegard ,

The idea of this enhancement is to support "in place processing of packets based on multiple checks".

As an example, see this:
```
tsp \
 -I file input.ts \
 -P filter --pcr -m \
 -P filter -o -p 1001 -m \
 -P filter -o -n --payload --in-place-process-remap 1002 \
 -P continuity -p 1002 -f \
 -P pmt -s 100 --pcr-pid 1002 \
 -O file output.ts
```

With this command you can change the PCR from the stream with pid 1001 to pid 1002 in the service 100, when the PCR is inside packets without payload (this is not a regular behaviour, but you can force your compressor the do it).

As you can note the parameters `-o` and `-m` are used to mark packets, and this enables the use of the AND criteria without filtering out any packet. Then the parameter `--in-place-process-*` does the process (you can add new candidates like "set-adaptation-field" or similar). And the parameter `--every` is useful if you just want to trigger the true every N occurrences (not used in this example, but useful in other environments even without these new parameters).

I hope you like to merge it.
Regards.

lelegard:
Hi,

Interesting idea indeed. I have not checked the code thoroughly, only had a look at the GitHub view of the commit. Does this mean that "marked" packets can go out of the `tsp` chain with `0xff` instead of `0x47`?

I am thinking of an even more generic kind of "packet labelling" inside `tsp`, without killing the sync byte of each packet.

lars18th:
Hi @lelegard ,

I'm glad you said that. I feel it's a nice but tricky change inside TSDuck.

Let me explain some questions:
- `--every`: This is not related to the rest. But I feel it's a must have for the `filter` plugin. Please, add this part of the code. The objective is quite simple and easy to understand.
- "Marked packets": This PR works for my objective. But I agree with a more generic implementation. What it does is to use the SYNC byte (0x47) to share information between modules. I'm inspired by your plugin loop control (`tspProcessorExecutor.cpp`), that sets `pkt->b[0] = 0;` for dropping a packet. As you can imagine, this will be transparent for filters, except if they check the SYNC byte.
- Regarding the use of marks with the SYNC byte I suggest a "bit flag" approach to maximize the use and support multiple marks:
```
0000 0001: DROP/PASS the packet (the same current use).
0000 0010: MARK1.
0000 0100: MARK2.
0000 1000: MARK3.
0001 0000: reserved.
0010 0000: reserved.
0100 0000: reserved.
1000 0000: SYNC byte valid (this indicates an original, and checked, 0x47 value). If not set indicates an spureous original value (!=0x47).
```
- Then many plugins can be enhanced to support parameters such as `-o 1|2|3` which indicates that it only applies to packets with these marks. And `-m 1|2|3` to the SET flag of marks, instead of dropping/erasing packets.
- `-P inplace-process`: Futhermore you can create then a new plugin that instead of "filter" it does "process". Then functions like `--remap`, `--set-pusi`, etc. can be implemented on it. However, for pid remapping if the `remap` plugin will supports flags then it will be irrelevant. And for others will be similar. For example, the `pattern` will be a good candidate to support flags. The top goal could be a plugin for execute some external process that receives the packet and sends a new content (it will be the generic packet processor). In this case what ss required is to support OR and AND checks, and this approach adds it.

I hope you will consider doing one of the two options (merge or enhance). But don't forget to merge the `--every` part.

Regards.

lelegard:
My concern is that, with your code, marked packets will be output (in the `-O` plugin) with invalid sync bytes, not `0x47`. The result is an invalid transport stream. We cannot do that.

Generally speaking, plugins shall process valid packets only. They should not make them invalid by messing with the sync byte.

The zero value is just an internal trick inside `tsp`. Once zero is set, the packet will no longer be presented to any plugin and will not be output.

I am thinking about an internal "per-packet metadata" mechanism inside `tsp` with a documented interface to plugins. I will let you know when I have more clear ideas.

lars18th:
> I am thinking about an internal "per-packet metadata" mechanism inside `tsp` with a documented interface to plugins. I will let you know when I have more clear ideas.

Great! I agree it's best to think on it before implementing anything. In any case I'm glad that you consider it.

Regards.

lelegard:
I have added a packet metadata mechanism in `tsp`. The price to pay is to change the API of all packet processing plugins. But the benefit is an evolutive way of propagating per-packet metadata from plugin to plugin. The metadata contain 32 _labels_ numbered from 0 to 31.

lelegard:
Just a remark:

> when the PCR is inside packets without payload (this is not a regular behaviour, but you can force your compressor the do it).

This is a perfectly regular behaviour. This is typically seen when a service has a dedicated PCR PID which is neither audio nor video. In that case, the PCR PID carries only PCR's. The bitrate is low and each packet contains a PCR but no payload. I have seen that several times.

lelegard:
I have implemented the packet metadata feature and I use them to implement _packet labelling_. Have a look at the options `--set-label` and `--label` in plugin `filter`. Also have a look at option `--only-label` which is now common to all packet processing plugins.

I will probably add a section about packet labelling in the documentation of `tsp`.

lelegard:
Hi @lars18th 

I have added the missing feature, option `--every`, from your description. As you mentioned, it is not related to the rest but can be useful and is trivial to implement.

All the rest of your proposal can be done using packet metadata and packet labelling. Moreover, packet metadata open the door for future features. For instance, it is used to implement options `--nullified` and `--input-stuffing`. I don't know if they are useful but they were trivial to implement and demonstrate the usage of packet metadata.

I have added the description of packet labelling in the reference section of `tsp`.

Please review this. Then, you may close this PR without merging because the final implementation took a different route for the same result.

lars18th:
Hi @lelegard ,

Very good work with your implementation! I'm very satisfied. :+1: 

However, before close this PR, two comments:

1. Regarding the plugin `filter` I'll open another issue. Please, review it.
2. Your implementation about `--every` is not correct (almost from my point of view). Let me to explain why:

- The original idea behind my implementaion is to "catch" only packets that _satisfy_ the check.  One example with "every=2": the first time that one packet satisfies the check we don't do anything with it; and when a *second* one satisfies the condition, in this case (and only in this case) the statement is executed. However, your current implementation count "incoming" packets, and not "ok" packets (I use the label "ok" as any packet that satisfies the condition).

I'm sure you get it. Therefore, in order not to change the current implementation (that has sense too), I suggest to add another new parameter to the plugin `filter`: `-every-ok` In this case the behaviour is the one described by me.

You agree?

lars18th:
Hi @lelegard ,

>  Therefore, in order not to change the current implementation (that has sense too), I suggest to add another new parameter to the plugin filter: -every-ok .

Don't worry. This isn't necessary. It's possible to implement it with another `filter` instance that uses this syntax: `-P filter --only-label X --every N --set-label Y`.
In any case please use this as example in the documentation. It's pretty clarifying.

So, I close the PR. ALL IS WORKING! :+1:

PR #307, opened 2019-05-28
tsByteBlock:  Deal with linking issue on Windows

alevinsn:


lelegard:
Hi @alevinsn 

Are you sure that this code works both in debug and non-debug mode, while using the ">>" operator from the application? I recall from the past that using TSDUCKDLL on out-of-class exported function was mandatory. Otherwise, there were cases wit link errors.

alevinsn:
I don't think so.  According to https://devblogs.microsoft.com/oldnewthing/?p=2123 , there is basically no benefit to declaring inline functions with dllexport/dllimport, and instead, there are more likely to be unintended consequences to doing so.  I have a plan to go through all functions declared inline that have TSDUCKDLL specified and remove the TSDUCKDLL specification, assuming this is a patch that you would accept.

If the << and >> operator functions were implemented in the .cpp file, then I think the TSDUCKDLL specification would be required.

lelegard:
The article is probably not entirely true. Specifically, it assumes that an inline exported function will never be called. But this is not true. It depends on compilation options. I cannot remember what was the problem and which version of MSVC it was. But the TSDUCKDLL directive was definitely added because of some linking issue. Maybe it no longer applies to VS 2017.

Anyway, I am not comfortable with this PR. First, it does not cover all cases, in theory: what about other ByteBlock methods and inherited ones, what about other classes using similar heritage from the STL such as UString? Second, I am not sure we have investigated all consequences on applications.

alevinsn:
The article mentions some cases in which an inline exported function will be called, when the compiler doesn't chose to inline a function, but that's only because it was exported.  The article states that, even if a function that is declared inline isn't inlined for some reason, it will be handled correctly anyway.  If it is exported, it will find the one that is exported--if not, then it will declare a single instance of this function for use in the entire binary.  But, if it is exported, there are consequences to doing so.  Regardless, I didn't remove the TSDUCKDLL directive from the inline functions to be able to build--I did so because I don't think they are necessary, and so it's not so important to me that this aspect of the patch make it in.  Also, I'm building using Visual Studio 2015, and it works fine with this patch.

I think I've responded over e-mail to your other points, but the patch covers all the ByteBlock methods, since those methods that are declared externally, in the .cpp file, are exported, and those that are entirely declared in the .h file are covered implicitly and don't need to be exported.

PR #312, opened 2019-05-31
Correct the maximum buffer size of a UTF-8 string

masnagam:
Code points between U+8000 and U+FFFF are encoded using 3 octets in
UTF-8.

### DIFF
--- a/src/libtsduck/tsUString.cpp
+++ b/src/libtsduck/tsUString.cpp
@@ -326,8 +326,8 @@ ts::UString& ts::UString::assignFromUTF8(const char* utf8, size_type count)
 
 void ts::UString::toUTF8(std::string& utf8) const
 {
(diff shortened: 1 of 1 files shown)


lelegard:
~~~
-    // The maximum number of UTF-8 bytes is 1.5 times the number of UTF-16 codes.
+    // The maximum number of UTF-8 bytes is 3 times the number of UTF-16 codes.
~~~

Yes, you are right. It is 1.5 time the _number of bytes_ in the UTF-16 string, hence 3 times the _number of characters_ in the the UTF-16 string.

I am a bit ashamed of this bug 😞  Thank you for noticing and fixing it. I will merge right away.

masnagam:
tsduck is great.
I really respect your efforts and contributors 🥇

Issue #321, opened 2019-06-14
Request: Merge-Tables plugin


lars18th:
Hi,

Based on the idea of #318 I do some DIY test merging 188 packets interlaced from two streams (with identical CBR). Then I marked them with label 1 and 2 using the regular packet position (easy: `1212121212...`).

However, now I need to apply the same functionality as the `merge` plugin, but this plugin only works with two streams: the main one and another from the standard input. So, in this scenario we need another plugin with similar (or identical functionalities) but based on pids/flags on the same TS.

I deep describe the scenario:
- Inside the same TS are 0-21 pids from two different streams.
- Each table is marked with an unique flag (1 or 2), so pid 0 from TS-1 has flag 1 and pid 0 from TS-2 has flag 2.
- Other pids (data) have flags too, but without collisions (unique pids for each substream). 
- Then the idea it's something like `tsp -I file input.ts -P merge-tables main-flag=1 merge-flag=2 -O file output.ts`.
- And the rest of the plugin options can be the same.

You think this has sense?
Or is there an alternative?

lars18th:
Hi,

More or less, I feel I can do that with something like:
```
tsp ... -P sections -n -s --pid 16 --pid 8016 -P sections -n -s --pid 18 --pid 8018 ...
```

And this "merges" the tables 16 (SDT) and 18 (EIT) with the alternatives at 8016 and 8018.

However, one simple question: With the plugin `sections` it's possible to accomplish the **entire** functionality of the `merge`plugin (regarding the PSI/SI tables)?

Thank you!

lars18th:
Hi @lelegard ,

As commented in the #325 it's easy to rename multiple sub-streams (original TS muxed manually inside the same TS). When every sub-stream has the same ID, then it's more easy to adapt the PSI/SI tables.

However, despite the EIT tables (we can use the `sections` plugin for this task) we don't have any method to combine (or join) multiple PAT, SDT/BAT or NIT tables. So I suggest a new plugin to do something like (sub-streams have labels 1 and 2):
```
tsp ... -P remap --only-label 2 -n -u 0-21=8000 \
 -P pat-join -p 8000 \
 -P nit-join -p 8016 \
 -P sdt-join -p 8017 \
 -P sections -n --pid 18 --pid 8018 \
 ...
```

[...]

Issue #318, opened 2019-06-13
Request: Interleave option for File input


lars18th:
Hi @lelegard ,

First of all, I know about the current `merge` and `inject` plugins. But I request for a very **simple** new plugin (I feel it's _very_ simple) for another type of merging/muxing/adding sources.

I think it's best to present it with an example:
```
tsp -I combine file1.ts file2.ts -P pcradjust -O file remux.ts
```

What's doing this `combine` plugin? It only takes 1 packet from file1 then 1 packet from file2, etc.
That's simple! For N input files it takes 1 packet of each, and if one file ends, it uses NULL stuffing to grant a constant bitrate until the last filter ends.

What's the use case?
- We want to remux (combine) two (or more) TS sources in a _deterministic_ way, without "injecting" a TS into another. This action will maintain the internal structure of the sources.
- So, we use other filters to first adapt the source files to **identical** CBR values. For example `add-input-stuffing` global parameter to increase the bitrate; `reduce` plugin to decrease the bitrate; and `limit` to cut-off the bitrate.
- In some cases, before the previous filtering we can remove some services, change tables, remap pids, etc.
- And after the "combination" we can do more postprocessing to resolve some conflits and recompute the bitrate.

In fact it's a simple "round-robin dumb remux" but called _combine_.
What you think?

lars18th:
Hi,

I forgot to mention an optional parameter: `--with-labels`. This parameter will set label 1 for packets from input 1, label 2 for input 2, and so on. Then it will be more easy to manipulate the packets after the muxing without preprocessing the inputs.

Thank you @lelegard for listening our crazy ideas! :wink:

lars18th:
Hi @lelegard ,

After thinking more on it, I see that:

1. It's an INPUT plugin, and not a processing one.
2. Two other parameters have sense: `--stop-at-first-end` and `--stop-at-last-end`. I suggest to use the last as the default. They are self explanatory (when using the second, NULL stuff is added when some input ends).
3. One _non_ trivial addition is to support not only files but other input types. However, I feel we don't need this at time.

I hope you will like to implement it.
Regards.

lelegard:
Yet another "crazy" idea indeed. I really wonder why anyone would need this.

A few remarks:

- I would prefer the name `interleave` which is more accurate (and more TVish) than `combine`.
- The best way would probably be to add some `--interleave` option to the input plugin `file`. It already accepts several files in the argument list and plays them sequentially. With this option, they would simply be interleaved.
[...]

Issue #139, opened 2018-10-19
svremove plugin optimization


lars18th:
Hi @lelegard ,

I comment this only for convenience.
Please *only* consider it if it's more efficient than a simple repetition of the plugin.

For the `svremove` plugin I suggest these to enhancements (a list of services, and the not operator):

- `tsp -P svremove 100,200,300,400`: This indicates to remove four services at the same time.
- `tsp -P svremove !100`: This instructs to remove _all_ services _except_ the service 100.
  
I feel it's best to do the work in one plugin instance, than repeat it a lot of times for the same objective (more resource intensive).

What you think?

lelegard:
Hi, @lars18th,

As you have experienced multiple times, the project is open to suggestions and enhancements. But this time, the answer is a firm "no" because this is against the basic principles of TSDuck: Keep It Simple and Stupid (KISS), implement basic and simple features, provide the ability to combine them in an infinite number of ways. 

If something can be done by using a plugin twice, I do not want to make the plugin more complex by modifying it to do it in one pass. Because if we do this once, we will have to do it everywhere. And the code will become more complex, which means more error-prone, more bugs, spending more time to maintain it. If TSDuck is so reliable after 10 years of maintenance and evolution, it is mainly because of such principles.

I accept to modify a plugin only when something cannot be done in two separate instances, when it needs to be done in once. 

The best example is the good old `grep` command. What do you do it you want to search for FOO _and_ BAR vs. FOO _or_ BAR? The _or_ must be implemented in the command: `grep -e FOO -e BAR` but the _and_ can be achieved with two instances: `grep FOO | grep BAR`. This is why the popular and universal `grep` command never implemented an _and_ in more than 40 years of existence. I apply the same principle.

lelegard:
And I forgot to address your second suggestion: `tsp -P svremove !100` is already implemented as `tsp -P zap 100`.

lars18th:
Hi @lelegard ,

After think more, I completly agree! Thank you for the good response, and for the tip of `tsp -P zap 100`.

I close this issue.

lars18th:
Hi @lelegard ,

I accept your well based opinion, and I don't want to resurrect an issue without a cause. But I have a problem:

- I need to remove _all_ services from a TS _except_ a short list (two or three). And I need to put it inside one script to process a list of TS that have an undefined number of services inside it. Any idea to do this?
```
tsp -P svremove --negate 100,200,300
```
[...]

PR #331, opened 2019-06-18
mux: fix incorrect initialization with inter-packet

lars18th:
When using the parameter "inter-packet" the initialization of the _pid_next_pkt is incorrect.

### DIFF
--- a/src/tsplugins/tsplugin_mux.cpp
+++ b/src/tsplugins/tsplugin_mux.cpp
@@ -281,6 +281,8 @@ ts::ProcessorPlugin::Status ts::MuxPlugin::processPacket(TSPacket& pkt, TSPacket
         }
         _inter_pkt = ts_bitrate / _bitrate;
         tsp->verbose(u"transport bitrate: %'d b/s, packet interval: %'d", {ts_bitrate, _inter_pkt});
+    } else if (_packet_count == 0 && _bitrate == 0) {
+        _pid_next_pkt = _inter_pkt;
     }
 
     // Count TS
@@ -365,7 +367,7 @@ ts::ProcessorPlugin::Status ts::MuxPlugin::processPacket(TSPacket& pkt, TSPacket
 
     _inserted_packet_count++;
     _pts_last_inserted = _youngest_pts;   // store pts of last insertion
-    tsp->debug(u"Inserting Packet at PTS: %'d, file: %s", { _pts_last_inserted,_file.getFileName() });
+    tsp->debug(u"[%d:%d] Inserting Packet at PTS: %'d (pos: %'d), file: %s (pos: %'d)", { _inter_pkt,_pid_next_pkt,_pts_last_inserted,_packet_count,_file.getFileName(),_inserted_packet_count });
 
     if (_inter_time != 0) {
         _pts_range_ok = false; // reset _pts_range_ok signal if inter_time is specified


lars18th:
Hi @lelegard ,

If you check something like `tsp -v --debug=3 -a 1/1 ... -P mux -i 2 ... `
then you will see (with the new debug line) that the muxed packets isn't inserted at the correct place.

Please, check it and merge this patch.
Regards.

lelegard:
Hi @lars18th 

I do not see the use of this modification. Apart from one debug message, the only modification is:
~~~
    } else if (_packet_count == 0 && _bitrate == 0) {
        _pid_next_pkt = _inter_pkt;
    }
~~~
But in `start()`, we already have, for all cases:
~~~
    _pid_next_pkt = 0;
~~~
and then after each insertion:
~~~
    _pid_next_pkt += _inter_pkt;
~~~
So, in all cases, the first packet is inserted at index zero (in fact, at the first null packet). The actual insertion point of the first packet is not really important. Only the repetition rate it.

I do not know if there is a bug in the plugin, but the modified code does not seem to fix anything.

lars18th:
Hi @lelegard ,

> I do not see the use of this modification.

Sorry, it seems to me you didn't understand the problem this time (with all respect). :wink: 

First, the new log info is required to see the problem. Please, compile only with the debug line of the patch and think on this example:

- `tsp -v --debug=3 -a 1/1 ... -P mux -i 2 ...`. Here you _double_ the input bitrate. So, every 1 packet you found a new null packet. Then with the `-P mux -i 2` you want to...
- Ejem! Before writing the answer, I copy from the documentation: `Specifies the packet interval for the inserted packets, that is to say the number of TS packets in the transport between two new packets.` 
- So, the _expected_ behaviour is that every 2 packets you insert one from the "muxed" file. Here a graphical example:
 -- After `-a 1/1`: `1n2n3n4n5n6n7n8n9n...`
 -- After `-P mux -i 2`: `1A2B3C4D5E6F7G8H9I...`
- Correct? I hope you agree.
- What's the problem? This **isn't** the obtained behaviour. Note this _other_ example:
 -- After `-a 1/1`: `1n2nnn4nnn6nnn8n9n...`  (note that the original stream includes null packets)
 -- After `-P mux -i 2`: `1A2BCD4EFG6HIJK8L9...`  :fearful: 
- Futhermore, the bug is worse, because when use `-P mux -i 3` (or a greater number) you obtain THE SAME BEHAVIOUR!

You can check all of this with the new debug line.

But regarding the new code:
```
    } else if (_packet_count == 0 && _bitrate == 0) {
        _pid_next_pkt = _inter_pkt;
    }
```

The objective of this code it's to initialize `_pid_next_pkt` not to ZERO! Why? Because if you initialize to 0 then you're inserting **not** at some interval from the _last_ insertion, but from the _last_ first counter. Please, note this code:

```
    if (_packet_count < _pid_next_pkt) {
        return TSP_OK;
    }
```

With an initialization with a zero value the check is every time true. This is because the `pid_next_pkt` is incremented by `_inter_pkt`, but the FIRST time you found a null packet you insert a new packet, even if it's the first packet because `_pid_next_pkt` is 0. Then in the next loop the value of the `_pid_next_pkt` will be incremented with `_inter_pkt`, but the `_packet_count` will be only increment plus 1. And as a consequency the check will continue to be true when a new null packet comes. So the problem is that the _packet_count it's always from behind `_pid_next_pkt`. Therefore, the correct initialization is to set the maximum value of a loop and this will guarantee the interval.

In any case, don't belive me: Please check with different values `-P mux -i 2`, `-P mux -i 3`, `-P mux -i 4`... using one TS that starts with multiple nulls. With the new debug line the bug will be evident!

Regards. :wink:

lelegard:
Hi @lars18th,

I have made some tests with the current version (without this PR) and the behaviour seems correct to me.

First, remember that the value from option `--inter-packet` is an average. In the general case, it is impossible to guarantee a strict interval since the insertion is opportunistic, replacing null packets when they arrive. So, in all tests below, if the average insertion rate is preserved, I consider it as fine.

First, I create 100-packet files, with null packets, PID 1 and 2:
~~~
$ tsp -I null 100 -O file null.ts
$ tsp -I craft --pid 1 -P until --packet 100 -O file 1.ts
$ tsp -I craft --pid 2 -P until --packet 100 -O file 2.ts
~~~

The base file is the one containing null packets. Now insert the file containing PID 1 with `-i 2`. You can see that the inserted packets come once every 2 packets as expected. There is an artifact in the first 2 packets, but it does not impact the average.
~~~
$ tsp -I file null.ts -P mux 1.ts -i 2 -O file n1n1n1.ts
$ tsdump n1n1n1.ts | grep PID:
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
~~~

Same test with `-i 3`, same conclusion:
~~~
$ tsp -I file null.ts -P mux 1.ts -i 3 -O file nn1nn1nn1.ts
$ tsdump nn1nn1nn1.ts | grep PID:
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
~~~

Another test using `nn1nn1nn1` as input and inserting the file with PID 2 with `-i 3`. Same conclusion:
~~~
$ tsp -I file nn1nn1nn1.ts -P mux 2.ts -i 3 | tsdump | grep PID:
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
~~~

You may say it is too easy because `nn1nn1nn1` and `-i 3` are synchronized. Let's try with `-i 2`.
~~~
$ tsp -I file nn1nn1nn1.ts -P mux 2.ts -i 2 | tsdump | grep PID:
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 8191 (0x1FFF), header size: 4, sync: 0x47
  PID: 1 (0x0001), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
  PID: 2 (0x0002), header size: 4, sync: 0x47
~~~

Visually, the result seems not so obvious. But remember that the insertion is opportunistic, on null packet replacement. So, let's display the total number of packets and you will see that there is approximately one PID 2 packet every 2 packets, as expected.
~~~
$ tsp -I file nn1nn1nn1.ts -P mux 2.ts -i 2 -P count -O drop
* count: PID    1 (0x0001):         34 packets
* count: PID    2 (0x0002):         51 packets
* count: PID 8191 (0x1FFF):         15 packets
$ 
~~~

My conclusion is that there is no need for a modification.

lars18th:
Hi @lelegard ,

First of all, thank you for doing all these tests. I appreciate you listening to me.
But...

> First, remember that the value from option `--inter-packet` is an average.

Regarding this, perhaps the problem is the perception about the sense of this parameter. If you think it's _average_ then the behaviour can be correct. However, in my opinion the sense requires to be _strict_. And what does it mean if the behavior is _strict_? From my opinion that the distance between the inserted packets _never_ be less than the interval value. Futhermore, in case of a new insertion can't be done in the _exact_ expected position (because a non-null packet is in this position) then you move _this_ packet, but no the rest.

> My conclusion is that there is no need for a modification.

Let me to explain how I found this bug, and perhaps with this example you will change your opinion:

- In order to accomplish what I described in #318, I found a different approach. The idea is quite simple: duplicate the bitrate of the input stream, and use the `mux` plugin to insert a second stream interlaced in the first stream. After that I have a _perfect_ muxed streams (interlaved 1-to-1), with the only requirement to be both CBR and with the same bitrate. After that applying the `reduce` plugin and `pcradjust` I can clean the output about any jitter problem. The only remaining problems are "merging tables" (issue #321) and remove unused services (issue #139). But with a lot of code, I can obtain some results.

- But I discovered a trouble with this technique: The second muxed TS will be out-of-sync after some packets. Then I writed the code of the new debug line (please, I strongly suggest to add almost this line to the code... It is only with it that we can clearly see the problem. For sure it doesn't change any behaviour... it's a simple enhanced debug line). And what the debug revealed was that "the mux interval is not respected", in the sense that time-to-time an additional packet is inserted. So after a lot of debug I discovered that the check `if (_packet_count < _pid_next_pkt)` works incorrectly as the second part is behaind the correct value. Just one loop behaind. And the solution is to initialize it to another value. After, I discovered that the correct value is the interval value minus one. However, that's equivalent to ` _pid_next_pkt = _inter_pkt` in the initilization phase.

So, after saying all this, I suggest we do the following, if that's all right with you:

1. You add the dumb new debug line to the code.
2. I prepare a new test that you can repeat and that shows the problem.

You agree?

lelegard:
> From my opinion that the distance between the inserted packets never be less than the interval value.

This is impossible, simply. If you specify `-i N` and there are sequences of N or more non-null packets, it is _impossible_ to have a strict interval. So, trying to get an average rate is the best we can do (and sometimes it is impossible if there is not enough stuffing overall).

You probably have a "file processing" standpoint. But broadcast does not work like that. Consider the complexity of a statistical multiplexing system - with all retro-action loops between the MUX and the encoders - and you will see that you cannot just do what you want with an existing stream in the general case.

lars18th:
Hi @lelegard ,

Thank you for your help. I really appreciate _all_ of your comments!

> > From my opinion that the distance between the inserted packets never be less than the interval value.
> 
> This is impossible, simply. If you specify `-i N` and there are sequences of N or more non-null packets, it is _impossible_ to have a strict interval. So, trying to get an average rate is the best we can do (and sometimes it is impossible if there is not enough stuffing overall).

That's the test that I'm preparing:
```
tsp -a 1/1 --debug=3 -I file ./test-003.ts \
 -P mux --no-pid-conflict-check --no-continuity-update -i 2 ./test-003.ts -O drop
```
(`test-003.ts` is yours https://github.com/tsduck/tsduck-test/blob/master/input/test-003.ts)

In this case the _strict_ interval is really feasible! Every two packets the second one is a null, becuase the `-a 1/1`. Even, if you use `-a 1/2` the use of `-P -i 3` can be strict as every 3 packets the last one will be a null. And so on. You can really generate this type of streams.

> You probably have a "file processing" standpoint. But broadcast does not work like that. Consider the complexity of a statistical multiplexing system - with all retro-action loops between the MUX and the encoders - and you will see that you cannot just do what you want with an existing stream in the general case.

Well, although my ultimate goal may be a statiscital multiplexing, this complexity is at time out-of-scope. However, in order to implement a good multeplexing I prefer my approximation. Now I can do it _only_ with files, but with the implementation of the Input with the enhancement commented at #318 it should be done in realtime too. Think on this example:

- You have three different TS.
- You can convert in realtime these TS to CBR with _identical_ bitrate (that can be done at time with current TSDuck tools).
- Then you can "interlace" them with a _perfect_ 3-to-1 mode, as 1-2-3-1-2-3-1-2-3... And that can be done in realtime _without_ any problem, as the sources are CBR with the same bitrate. The only problem could be a jitter problem, but this can be solved with a simple buffer (included at time in TSDuck).
- - Now I emulate this task with `-a 2/1` and two `-P mux -i 3`.
- After this "muxing" you can modify/edit PSI/SI tables and I remove unused stuff, for example with multiple `-P svremove` plugins.
- And the last task to complete a _perfect_ (in the sense of "good for broadcast") is `-P reduce XXX YYY -P pcradjust --bitrate NNNN --ignore-dts --ignore-pts`. And this _really_ works and can be done in realtime.

So in fact, this type of muxing can be done in realtime. I'm not doing anything imposible. I check it using a modulator and it works with all receptors. So, please, have confidence with this approach.

Regards.

lars18th:
Hi @lelegard ,

I'm closing this PR because I think the problem is gone. Futhermore now I prefer to use the "--interleave" mode of the Input file plugin.

Regards.

PR #334, opened 2019-06-20
Add offset-pattern option

lars18th:
Now when using the payload-pattern it's possible to define an offset with the parameter --offset-pattern.


lars18th:
Hi @lelegard ,

Based on parameters `--offset-pusi` and `--offset-non-pusi` of the plugin `pattern`.
Useful to replace the content of a PES packet without disturbing PTS/DTS marks.

I hope you agree to merge it!

lars18th:
Hi @lelegard ,

I included the option `--pes-payload` too for referencing the PES payload instead of the TS payload.
With this update will be easy to replace the start of a PES packet without touching PCR, DTS and PTS marks.

Regards.

Issue #378, opened 2019-07-24
Compilation error: Debian 7 GCC-4.9-backport


lars18th:
Hi @lelegard ,

Compiling in a Linux Debian 7.11 server 64 bit, with the GCC-4.9-backport compiler installed, I see this error:
```
  [CC] release-x86_64/tsduck.so
  [AR] release-x86_64/libtsduck.a
/usr/bin/ar: invalid option -- 'U'
Usage: /usr/bin/ar [emulation options] [-]{dmpqrstx}[abcDfilMNoPsSTuvV] [--plugin <name>] [member-name] [count] archive-file file...
       /usr/bin/ar -M [<mri-script]
 commands:
  d            - delete file(s) from the archive
  m[ab]        - move file(s) in the archive
  p            - print file(s) found in the archive
  q[f]         - quick append file(s) to the archive
  r[ab][f][u]  - replace existing or insert new file(s) into the archive
  s            - act as ranlib
  t            - display contents of archive
  x[o]         - extract file(s) from the archive
 command specific modifiers:
  [a]          - put file(s) after [member-name]
  [b]          - put file(s) before [member-name] (same as [i])
  [D]          - use zero for timestamps and uids/gids
  [N]          - use instance [count] of name
  [f]          - truncate inserted file names
  [P]          - use full path names when matching
  [o]          - preserve original dates
  [u]          - only replace files that are newer than current archive contents
 generic modifiers:
  [c]          - do not warn if the library had to be created
  [s]          - create an archive index (cf. ranlib)
  [S]          - do not build a symbol table
  [T]          - make a thin archive
  [v]          - be verbose
  [V]          - display the version number
  @<file>      - read options from <file>
  --target=BFDNAME - specify the target object format as BFDNAME
 optional:
  --plugin <p> - load the specified plugin
 emulation options:
  No emulation specific options
/usr/bin/ar: supported targets: elf64-x86-64 elf32-i386 elf32-x86-64 a.out-i386-linux pei-i386 pei-x86-64 elf64-l1om elf64-k1om elf64-little elf64-big elf32-little elf32-big plugin srec symbolsrec verilog tekhex binary ihex
make[2]: *** [release-x86_64/libtsduck.a] Error 1
```

The command line used to compile is:
```
[...]

PR #379, opened 2019-07-24
Fix compilation on some Linux machines

lars18th:
Fixes the compilation error with the AR tool described in https://github.com/tsduck/tsduck/issues/378


lars18th:
Hi @lelegard ,

Great to see you again. :wink: 
Any reason to not merge this plugin? I really need it in some Debian 7 machines.

lelegard:
Hi @lars18th,

I just quickly merged the "obvious" PR's before looking at more problematic ones. 

As a general rule, I want to understand an error before fixing it. Typically, changing something just because it suddenly works without understanding _why_ it now works is something I avoid to do.

Your PR replaces this:
~~~
ARFLAGS  = rc$(if $(MACOS),,U) $(ARFLAGS_EXTRA)
~~~
with this:
~~~
ifdef MACOS
    ARFLAGS  = rcU $(ARFLAGS_EXTRA)
else
    ARFLAGS  = rc  $(ARFLAGS_EXTRA)
endif
~~~

If I understand correctly your issue #378, the option `U` is passed to `ar`. I do not understand why `$(if $(MACOS),,U)` is replaced by `U` on a Debian system where `MACOS` is undefined. I want to understand this first. But Debian 7 is very old. I have no idea on the versions of the utilities which are present. This looks like a `make` issue. Which version of `make` is there? Why the `$(if ...)` expression should be incorrectly interpreted with this version of `make`?

lars18th:
Hi @lelegard ,

> I just quickly merged the "obvious" PR's before looking at more problematic ones.
> 
> As a general rule, I want to understand an error before fixing it. Typically, changing something just because it suddenly works without understanding _why_ it now works is something I avoid to do.

OK. I understand, no objection. :wink: 

> If I understand correctly your issue #378, the option `U` is passed to `ar`. I do not understand why `$(if $(MACOS),,U)` is replaced by `U` on a Debian system where `MACOS` is undefined. I want to understand this first. But Debian 7 is very old. I have no idea on the versions of the utilities which are present. This looks like a `make` issue. Which version of `make` is there? Why the `$(if ...)` expression should be incorrectly interpreted with this version of `make`?

Yes, the main problem is that the `U` option is pased to the `ar` tool.

However, I don't know the cause. After a lot of wasted time I discovered the error. And I can only provide this solution. In any case, please consider that (if you find it acceptable):

- A Debian 7 is not a "very" old distribution. A lot of running systems use it. Therefore, eliminating compatibility with it only by a simple bug doesn't seem to be reasonable.
- The problem seems to be more related with the _syntax_ used than with the tools. Is it really a big change to use the simple syntax of `ifdef ... else ... endif` instead of the complex expansion `$(if ...)`?
- I'm not sure, but perhaps in some other old distributions it works because the `ar` tool discards the unknown `U` option.

So, even if we don't discover the root cause I suggest to merge the patch.
In any case your decision will surely be correct, as it usually is.
Regards.

lelegard:
> I'm not sure, but perhaps in some other old distributions it works because the ar tool discards the unknown U option.

Not at all, the U option is _never_ passed to `ar` on any version of Linux. This is the purpose of the expression `rc$(if $(MACOS),,U)`. It translates to `rcU` when `MACOS` is defined and non-empty (ie. on macOS). And it translates to `rc` only when `MACOS` is undefined or empty (ie. on Linux).

This is why it is weird that `rc$(if $(MACOS),,U)` translates to `rcU` on Debian. There is a hidden problem that must be understood.

Among the possibilities:

- The version of `make` is buggy and `$(if ..)` structures are incorrectly translated. This is serious because such structures are used elsewhere. With `ar`, you see it because it passes an invalid option on Linux but in other cases, this may cause a silent but invalid configuration. In that case, that version of `make` should be replaced by a non-buggy one.
- The configuration of your system is incorrect for some reason and this configuration should be fixed.

In both cases, the correct solution is not to replace the `$(if...)` by `ifdef...`.

lars18th:
Hi @lelegard ,

> * The version of `make` is buggy and `$(if ..)` structures are incorrectly translated. This is serious because such structures are used elsewhere. With `ar`, you see it because it passes an invalid option on Linux but in other cases, this may cause a silent but invalid configuration. In that case, that version of `make` should be replaced by a non-buggy one.

OK. Now I understand the key question to arrive at the root cause.

> Not at all, the U option is _never_ passed to `ar` on any version of Linux. This is the purpose of the expression `rc$(if $(MACOS),,U)`. It translates to `rcU` when `MACOS` is defined and non-empty (ie. on macOS). And it translates to `rc` only when `MACOS` is undefined or empty (ie. on Linux).

Ummm... My assumption goes in another direction. The phrase "is never passed" can be false. It should be more like "it should never be passed", but: we are sure that _never_ is passed on all Linux and some `ar` tools discard the `U` option? Are you really sure about this? One thing is what you want to run, and another is what the system runs.

> This is why it is weird that `rc$(if $(MACOS),,U)` translates to `rcU` on Debian. There is a hidden problem that must be understood.

OK. I'll try to found the root cause.

> Among the possibilities:
> 
> * The version of `make` is buggy and `$(if ..)` structures are incorrectly translated. This is serious because such structures are used elsewhere. With `ar`, you see it because it passes an invalid option on Linux but in other cases, this may cause a silent but invalid configuration. In that case, that version of `make` should be replaced by a non-buggy one.

Perhaps... I'll explore it.

> * The configuration of your system is incorrect for some reason and this configuration should be fixed.

I don't think so.

> In both cases, the correct solution is not to replace the `$(if...)` by `ifdef...`.

OK. You're the boss. :wink:

lars18th:
Hi,

In the Linux where this patch is required:
- `make -v`: `GNU Make 3.81`
- `/usr/lib/gcc-4.9-backport/bin/gcc-ar --version`: `GNU ar (GNU Binutils for Debian) 2.22`
- `cat /etc/debian_version`: `7.11`
- `uname -a`: `Linux server 3.2.0-4-amd64 #1 SMP Debian 3.2.88-1 x86_64 GNU/Linux`

Any idea for a test?

lars18th:
Hi @lelegard ,

After reading the documentation of the GNU make https://www.gnu.org/software/make/manual/make.html#Conditional-Functions

I arrive to the conclusion that the problem is this:
- Instead of `ARFLAGS  = rc$(if $(MACOS),,U) $(ARFLAGS_EXTRA)`
- The correct syntax should be `ARFLAGS  = rc$(if $(MACOS),U) $(ARFLAGS_EXTRA)`

I'm right?

lars18th:
Hi @lelegard ,

I confirm that with a single comma instead of two the bug is fixed. Now I can compile in that Debian 7 server without problems.

It seems to me that nobody had detected this error before because the `ar` tools used didn't care about the wrong `U` parameter.

I hope you want to merge it now!
Regards.

lelegard:
My apologies, vacations under the sun must have burnt my brain even more than my skin 😕 

Yes, of course, `rc$(if $(MACOS),,U)` translates to `rc` on macOS and `rcU` on Linux. This was done on purpose because the option U does not exist on macOS.

I was confused because your fix does the opposite and I blindly followed that interpretation. So, first, this demonstrates that your fix cannot be used because it would break the macOS build.

Now, why is option U mandatory on Linux. See `man ar`:
~~~
D   Operate in deterministic mode.  When adding files and the archive index use zero for UIDs, GIDs, timestamps, and use
    consistent file modes for all files.  When this option is used, if ar is used with identical options and identical
    input files, multiple runs will create identical output files regardless of the input files' owners, groups, file
    modes, or modification times.

    If binutils was configured with --enable-deterministic-archives, then this mode is on by default.  It can be disabled
    with the U modifier, below.
...
U   Do not operate in deterministic mode.  This is the inverse of the D modifier, above: added files and the archive
    index will get their actual UID, GID, timestamp, and file mode values.

    This is the default unless binutils was configured with --enable-deterministic-archives.
~~~

In the past, I ran into situations and distros (can't remember which one however) where `ar` was configured with `--enable-deterministic-archives`. In that case, the timestamps in the archives are always zero (as with option D) and `make` will always rebuilds the archive, all the time, even if nothing is recompiled, because the timestamp of each module in the archive is lower than the corresponding object file. To fix that situation, the option U was used to force the actual timestamp of the object files to be used in the archive, regardless of the `ar` default mode.

On macOS (which is a BSD system, not a GNU one), the `ar` command does not have deterministic mode, no option D or U. This is the reason for `rc$(if $(MACOS),,U)`.

Now, the root cause of your problem is: why does your `ar` reject the option U? Probably because your system is too old.

However, on a CentOS 7.6 (the oldest Linux system I have), with GCC 4.8.5 (very outdated), `ar` comes with the package `binutils` version 2.27. By contrast, on a very recent Fedora 30, `binutils` has version 2.31.

Probably, some old version of the GNU `ar` did not implement the "deterministic" mode and the options D and U were unknown (as with the BSD `ar` which is used on macOS).

What is your version of `binutils`? Use `ar --version` on the base system and also `/usr/lib/gcc-4.9-backport/bin/gcc-ar --version`.

If there is no other solution, simply add `ARFLAGS=rc` to your `make` command, after `AR=/usr/lib/gcc-4.9-backport/bin/gcc-ar`.

This seems to be the proper fix to me: add another specific configuration on a very old system but do not modify `Makefile.common`.

If you agree on this, then you should close the PR without merging.

lars18th:
Hi @lelegard ,

> My apologies, vacations under the sun must have burnt my brain even more than my skin 

:stuck_out_tongue_winking_eye: 

> If you agree on this, then you should close the PR without merging.

I prefer to fix this bug in some way. I put here the requested info:

```
$ /usr/lib/gcc-4.9-backport/bin/gcc-ar --version
GNU ar (GNU Binutils for Debian) 2.22
Copyright 2011 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) any later version.
This program has absolutely no warranty.
```

And I confirm that with `--help` the options `-D` and `-U` are missing in this `ar` tool.
Futhermore, the system `ar` tool is the same (identical version) as with the backport of GCC-4.9.

So, my suggestion is this:
- As the MacOS also requires to *not* use the `-U` option (because is BSD based), then it's good to maintain the compatibility with environments that doesn't support this option. Other BSD-based or embedded systems may behave identically requirements.
- So, I suggest to add some "runtime selection" of the `-U` option based on a test. For example, checking the tool before using it.

What you think?

lelegard:
Version 2.22 is indeed very old for binutils.

> So, I suggest to add some "runtime selection" of the -U option based on a test. For example, checking the tool before using it.

But on which criteria? When you install a specific GCC 4.9 on Debian 10, the `ar` which comes with it understands option U. But when you install a specific GCC 4.9 on Debian 7, the `ar` which comes with it does not. What would you test? The `ar` file name? Parse its output with `--version`? Too complicated, probably unreliable (many other non-standard configurations will emerge with other users).

The macOS case is different. This is a BSD system, there is no U option, even on the latest versions. This is a _nominal case_ which must be handled transparently in the makefile.

On Linux, with reasonably recent versions (ie. less than a few years old), the nominal case is also transparently handled in the makefile.

In your case, you have a very old version. GCC is outdated, binutils is outdated, etc. By making some effort (installing a backported GCC and specifying `make` options), you can compile. That is fine for you. But this is an old and unsupported version which is not a nominal case and not transparently handled in the makefile.

You have to specify `GCC=...` and `AR=...`. You also need `ARFLAGS=...`. This is consistent.

Transparently handling outdated `ar` but not outdated `gcc` is not consistent.

You have an outdated system with non-standard tools (the backported GCC). You need a specific command or script to build TSDuck. And this command needs `ARFLAGS=...`. That's life...

lars18th:
Hi @lelegard ,

> You have to specify `GCC=...` and `AR=...`. You also need `ARFLAGS=...`. This is consistent.

If using `make ... ARFLAGS=rc ...` works, then I agree with you. I'll try it.
Regards.

lars18th:
Hi @lelegard ,

I complete the test, and you're right (as usual).
So, my final recommendation is this:
- Update the documentation to reflect this potential error, and how to overcome it.

So, please if you agree update the file https://tsduck.io/doxy/building.html#reqraspbian adding something like:
_If you see an error regarding the `ar` tool, try to compile using the command `make ARFLAGS=rc`. This have special relevance if you use old binutils package (version < 2.28) and a backport compiler (i.e. GCC 4.9)_

I hope you agree with it!
Regards.

lars18th:
Hi @lelegard ,

Finally I missed some point related to this (the compilation of the TSDuck in a legacy environment):
- When using a backport compiler that uses _a non standard path_ (for example the gcc-4.9-backport uses for libs: `/usr/lib/gcc-4.9-backport/lib` it's required to edit the file `<TSDUCK>/src/tstools/release-x86_64/setenv.sh` and add at the end of the second line `export LD_LIBRARY_PATH=[...]:/usr/lib/gcc-4.9-backport/lib`.

I suggest to _almost_ document it. Or if you agree add some support in the Makefile with a command like: `make LD_LIBRARY_PATH=/usr/lib/gcc-4.9-backport/lib` and this will _add_ (to the `release-x86_64` directory) the additional libs directory in the file `setenv.sh`.

What do you think?

lars18th:
Hi @lelegard ,

Thank you for the commit df968cfd17552ad461e17c7ae8c8b32964b7834e. :+1: 
So I close this PR.

PR #390, opened 2019-08-02
Various enhancements to DektecOutputPlugin

alevinsn:


lelegard:
Hi @alevinsn 

Thank you for this contribution. It looks interesting. I never encountered the kind of problem you describe in the comments but I understand the idea. I will not be able to test it since I no longer use ASI devices. BTW, this is funny because the very first reason to develop the first version of TSDuck, back in 2005, was stream processing between an ASI input and an ASI output.

Anyway, I trust your code. I was about to release version 3.18 tonight but I will integrate your contribution first. I will release the version over the week-end.

And thank you for fixing the typos in the help text. I will propagate them and your new options in the user's guide.

PR #396, opened 2019-08-26
Bugfix: section_syntax_indicator must be ignored for Stuffing Table

SS-FF:
**The stuffing table uses a short header but allows arbitrary values for the section_syntax_indicator:**
                           

image


lelegard:
Hi @SS-FF 

I understand the idea. I identified this point a long time ago. My concern was that this is one of the very few points where MPEG and DVB are incompatible. ISO/IEC 1381861 / H.222 is very clear about the "private section" syntax (private section meaning a table id in range 0x40-0xFE). If the section_syntax_indicator is 1, then the syntax must be a "long section" one.

In itself, the DVB stuffing section is useless. There are MPEG-defined ways to introduce stuffing, either at TS level or inside an elementary stream carrying sections.

So, why defining something that is useless? I think that the original idea behind the DVB stuffing table was the ability to easily hide or cancel a section "on the fly", at minimal cost. This can be done by patching the table id byte with 0x72, effectively turning the section into a DVB-defined stuffing section. But the rest of the section retains its original content and remains either a long or short section, depending on the section_syntax_indicator.

I agree that this is only speculation. But this is the only way to remain compatible with MPEG, which should be the unquestionable basis under DVB. This is why I deliberately continue to consider a DVB stuffing section as polymorphic, being either a long or short section, based on the section_syntax_indicator.

I understand that this is not exactly what DVB says in ETSI EN 300 468. But, when two standards conflict, the more fundamental one (here MPEG) should take precedence. Especially if an "interpretation" of the least fundamental one (here DVB) makes it compatible with the more fundamental one.

Did you run into some actual problems with DVB stuffing tables and section_syntax_indicator ?

SS-FF:
> Did you run into some actual problems with DVB stuffing tables and section_syntax_indicator ?

Yes, that's how I came to this.

I had a stream from a sat transponder where stuffing table sections were sent on PID 16 (SDT).
Those sections had the section_syntax_indicator indicating a long header.
But that's clearly not the case for stuffing tables so I looked it up and saw that the section_syntax_indicator should be disregarded for stuffing tables.

The result of this was that the CRC check is failing and since I've added some debug reports for failing CRC checks, I got aware of this.

At first I thought about disabling CRC checking for stuffing tables, but I found it more correct to avoid considering those sections as having a long header in the first place.

I do have a recorded stream - in case you'd like to see yourself. I just don't know where to upload.

lelegard:
> stuffing table sections were sent on PID 16 (SDT).

PID 16 is for the NIT. PID 17 is for SDT and BAT. 

> I do have a recorded stream - in case you'd like to see yourself. I just don't know where to upload.

You may simply extract PID 16 and keep a few hundredths packets and attach the result to a post here ("`tsp -I file in.ts -P filter --pid 16 -P until --packet 100 -O file out.ts`")

It would be interested to patch the table id with 0x40 and 0x41 (ID 16) or 0x42, 0x46, 0x4A (PID 17) and check if the CRC becomes correct. Also use `tstables` on the patched table when the CRC is correct. This could validate or invalidate the speculated use case for the stuffing section.

SS-FF:
It's been late...of course it was NIT.

Here's the filtered stream:

[TransponderSendingStuffingTableSectionsOnPid16_Filtered.zip](https://github.com/tsduck/tsduck/files/3546694/TransponderSendingStuffingTableSectionsOnPid16_Filtered.zip)

lelegard:
Hi @SS-FF,

This is an extremely strange way of packetising sections. Just have a look at a pair of adjacent packets:
~~~
* Packet 17
  ---- TS Header ----
  PID: 16 (0x0010), header size: 4, sync: 0x47
  Error: 0, unit start: 1, priority: 0
  Scrambling: 0, continuity counter: 14
  Adaptation field: no (0 bytes), payload: yes (184 bytes)
  ---- Full TS Packet Content ----
  0000:  47 40 10 1E 00 72 F0 B4 FF FF FF FF FF FF FF FF  G@...r..........
  0010:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0020:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0030:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0040:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0050:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0060:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0070:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0080:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0090:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  00A0:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  00B0:  FF FF FF FF FF FF FF FF FF FF FF FF              ............

* Packet 18
  ---- TS Header ----
  PID: 16 (0x0010), header size: 4, sync: 0x47
  Error: 0, unit start: 1, priority: 0
  Scrambling: 0, continuity counter: 15
  Adaptation field: no (0 bytes), payload: yes (184 bytes)
  ---- Full TS Packet Content ----
  0000:  47 40 10 1F 5D FF FF FF FF FF FF FF FF FF FF FF  G@..]...........
  0010:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0020:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0030:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0040:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0050:  FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF  ................
  0060:  FF FF 40 F3 EF 00 01 D5 00 02 F0 78 40 08 41 53  ..@........x@.AS
  0070:  54 52 41 20 31 00 4A 07 04 54 00 01 00 0C 01 4A  TRA 1.J..T.....J
  0080:  07 04 57 00 01 00 0D 01 4A 07 04 57 00 01 00 00  ..W.....J..W....
  0090:  04 4A 09 04 54 00 01 00 00 04 4E 4F 4A 08 04 57  .J..T.....NOJ..W
  00A0:  00 01 00 00 0A 02 5F 04 00 00 00 01 4A 07 04 57  ......_.....J..W
  00B0:  00 01 1E E6 90 4A 07 03 EF 00 01 13              .....J......
~~~
The first packet starts with a stuffing section which spans over two packets and where absolutely everything but the 8-bit table id and 12-bit section length is overwritten by FFFFFF...

Then, the NIT starts in the middle of the second packet, at offset 0x62, right after the stuffing section.

There is no rational reason to do this, except overwriting some previous section (maybe some NIT-Other you do not want to expose). The cancellation by overwriting everything with FFFFF, no even preserving the long header or pretending it is a short header, is a bit rude. The result is an invalid MPEG section (regardless of DVB). Anyway, the resulting section will be ignored, either because it is invalid in the first place, or if you figure out it is a DVB stuffing section and, as such, it is useless.

Coming back to your original PR, I am a bit reluctant to implement some invalid MPEG processing to handle a DVB case which may violate the fundamental standard and is useless by definition (because this is stuffing).

SS-FF:
Well, sure - the DVB spec defines something here that does not conform to it's parent spec.
But we cannot sue the DVB spec writers and even if we could - it wouldn't change reality.

From a logical or a legal point of view it is clear that the parent spec should take precedence.
But from a pragmatic and practical point of view: 
- I dare to say that there will hardly be any case in the world where an ffmpeg stream would contain packets on a **PID 16** stream that are having a correctly formatted **section header** and indicating the **stuffing table id in** its header  - WHERE- 
- those packets would NOT be DVB packets but some other kind of private data not related to DVB

That will probably never happen - at least not when the scope is TV reception anyway.
For those reasons, I think that the DVB spec should have precedence here.

But I got no stakes in that PR. Actually I'm doing all parsing in a custom C# implementation. I just recently reworked our table parsing based on the code flow from SectionDemux which is clearly superior to what we had before.

The sole motivation for creating this PR (and some others that will follow) was about giving at least a few small bits in return for this exceptional library that you're sharing.

Feel free to do whatever you want do with this PR.  The case is already closed on my side..

PR #497, opened 2020-01-06
Fix wrong INT32 max value

spnngl:


lelegard:
This was a really stupid one !
Thanks for noticing and fixing it.

PR #500, opened 2020-01-13
I/O plugins for SRT protocol

spnngl:
Hi,

I made those IO plugins using [libsrt](https://github.com/Haivision/srt).
I still have issue for the non-blocking mode: tspInputExecutor.cpp:90 consider it an error if we return 0 on reception which is often the case for non-blocking. Can I change that behaviour or find another way ?
I'm available to make changes if needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@
 #  - NOCURL  : No HTTP support, remove dependency to libcurl.
 #  - NOPCSC  : No smartcard support, remove dependency to pcsc-lite.
 #  - NOTELETEXT : No Teletext support, remove teletext handling code.
+#  - NOSRT : No SRT support, remove "Secure Reliable Transport" code.
 #
 #-----------------------------------------------------------------------------
 
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@
 #  - NOCURL  : No HTTP support, remove dependency to libcurl.
 #  - NOPCSC  : No smartcard support, remove dependency to pcsc-lite.
 #  - NOTELETEXT : No Teletext support, remove teletext handling code.
-#  - NOSRT : No SRT support, remove "Secure Reliable Transport" code.
+#  - SRT : SRT support, add "Secure Reliable Transport" code.
 #
 #-----------------------------------------------------------------------------
 
--- a/Makefile.tsduck
+++ b/Makefile.tsduck
@@ -104,6 +104,12 @@ ifdef NOTELETEXT
     CFLAGS_INCLUDES += -DTS_NOTELETEXT=1
 endif
 
+ifdef NOSRT
+       CFLAGS_INCLUDES += -DTS_NOSRT=1
+else
+       LDLIBS := -lsrt $(LDLIBS)
+endif
+
 ifdef NOCURL
     CFLAGS_INCLUDES += -DTS_NO_CURL=1
 else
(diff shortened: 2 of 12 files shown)


lelegard:
What is the rationale behind non-blocking mode in an input plugin?

In the general case, non-blocking input is useful in a context where a given thread may want to do something else when no input is available immediately. This is typically essential in a _reactor_ design pattern.

The other model is the heavily multi-threaded architecture where each thread is dedicated to do one single thing, for instance input. This is the model which is used in TSDuck. In that case, non-blocking input is useless. The input thread has nothing else to do but waiting for input. 

What would you expect the input plugin thread to do when no input is available?

lelegard:
Also, you may have noticed that all CI checks failed because using libsrt adds some dependencies.

- For Fedora, the package is `srt-devel`.
- For Red Hat / CentOS, I have not found any suitable package.
- For Ubuntu, the package is `libsrt-dev`.
- For macOS, I haven't found any Homebrew package.
- On Windows, how would you automate the installation of libsrt?

So, maybe we have to disable SRT by default on Read Hat, CentOS and macOS. We need to find an automation solution for Windows. Having a feature only on Fedora and Ubuntu is not appropriate.

spnngl:
You're right, I removed all nonblock code.
Also I disabled SRT by default everywhere, the packages on linux distro and installation instruction for windows are on the [readme](https://github.com/Haivision/srt/blob/master/README.md).

lelegard:
> installation instruction for windows are on the readme.

Well, everything is manual here, no automation 😞 

Moreover, everything is "linuxish" and needs all kinds of ports of "linuxish" packages to Windows: "LibreSSL for Windows", "pthread for Windows", "cmake for Windows", etc.

This is typically the kind of things I have always avoided with TDuck. TSDuck equally uses SSL, threads and build automation. But it uses the native tools and libraries of the OS, libcurl and OpenSSL on Linux, WinInet on Windows for instance. This offers the best and smoothest integration, guaranteed of support, ease of build, etc.

For instance, how could we automate the setup of a Windows system to validate the build? Currently, starting from a bare Windows development system, we just need to install the Dektec SDK and NSIS (to build the TSDuck installer). See [here](https://github.com/tsduck/tsduck/blob/master/.github/workflows/nightly-build.yml#L42). I see no easy way to install an SDK for libsrt. So, we cannot validate the build with libsrt.

Do not misinterpret me, I am not a Windows fan and I do prefer Linux for technical tasks such as everything that TSDuck does. But Windows users exist (unfortunately...😃) and we need to provide good and reliable support.

spnngl:
Okay, too bad.
I do not have windows at work nor home, so cannot help for this system... I thought about adding `#if !defined(TS_WINDOWS)` :angel:

lelegard:
OK @spnngl, considering the amount of energy you spent on this, I am integrating this PR. It will be available only when libsrt is present at compile time. I found that this is on Ubuntu 19.04 onwards, Fedora 31 onwards and macOS.

I still have to apply some modifications so that these dependencies are nicely and automatically handled.

I also had to make some modification for Ubuntu where libsrt is 1.3.x. Some socket options are not defined. So, I had to add some ifdef on libsrt version. On Fedora, libsrt is 1.4 and the code compiles OK.

After I push the integrated PR in the master branch, please test it to make sure I did not break any of your intentions.

Thanks for you work.

lelegard:
I also updated the TSDuck user's guide from the `--help` of your plugin.

PR #505, opened 2020-01-15
Small libsrt changes

spnngl:
I made two small changes:
* always set messageapi option. It is true by default and we need to be able to turn it off if we want.
* check for locally compiled/installed libsrt in /usr/local/include for NOSRT setting

### DIFF
--- a/Makefile.tsduck
+++ b/Makefile.tsduck
@@ -106,7 +106,7 @@ endif
 
 ifeq ($(MACOS)$(NOSRT),)
     # Not on macOS and SRT not disabled, check if libsrt is available.
-    NOSRT = $(if $(wildcard /usr/include/srt/*.h),,true)
+    NOSRT = $(if $(wildcard /usr/include/srt/*.h)$(wildcard /usr/local/include/srt/*.h),,true)
 endif
 
 ifneq ($(NOSRT),)
--- a/src/libtsduck/base/network/tsSRTSocket.cpp
+++ b/src/libtsduck/base/network/tsSRTSocket.cpp
@@ -450,7 +450,7 @@ bool ts::SRTSocket::setSockOptPre(ts::Report& report)
 
     if ((_mode != SRTSocketMode::CALLER && !setSockOpt(SRTO_SENDER, "SRTO_SENDER", &yes, sizeof(yes), report)) ||
         (_transtype != SRTT_INVALID && !setSockOpt(SRTO_TRANSTYPE, "SRTO_TRANSTYPE", &_transtype, sizeof(_transtype), report)) ||
-        (_messageapi && !setSockOpt(SRTO_MESSAGEAPI, "SRTO_MESSAGEAPI", &msgapi, sizeof(msgapi), report)) ||
+        (!setSockOpt(SRTO_MESSAGEAPI, "SRTO_MESSAGEAPI", &msgapi, sizeof(msgapi), report)) ||
         (_conn_timeout >= 0 && !setSockOpt(SRTO_CONNTIMEO, "SRTO_CONNTIMEO", &_conn_timeout, sizeof(_conn_timeout), report)) ||
         (_ffs > 0 && !setSockOpt(SRTO_FC, "SRTO_FC", &_ffs, sizeof(_ffs), report)) ||
         (_iptos >= 0 && !setSockOpt(SRTO_IPTOS, "SRTO_IPTOS", &_iptos, sizeof(_iptos), report)) ||
--- a/src/libtsduck/tsVersion.h
+++ b/src/libtsduck/tsVersion.h
@@ -44,4 +44,4 @@
 //!
 //! TSDuck commit number (automatically updated by Git hooks).
 //!
-#define TS_COMMIT 1604
+#define TS_COMMIT 1606
--- a/src/libtsduck/tsVersion.h
+++ b/src/libtsduck/tsVersion.h
@@ -44,4 +44,4 @@
 //!
 //! TSDuck commit number (automatically updated by Git hooks).
 //!
-#define TS_COMMIT 1606
+#define TS_COMMIT 1608
--- a/src/libtsduck/tsduck.mk
+++ b/src/libtsduck/tsduck.mk
@@ -57,7 +57,7 @@ ifdef TS_MAC
     endif
 else
     TS_INCLUDES += -I/usr/include/PCSC -I$(TS_INCLUDE_DIR)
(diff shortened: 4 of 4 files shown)


lelegard:
One issue: Please remove the space between the two `$(wildcard ...)` in the two makefiles. With a space in between, the result is never empty and SRT is always used, resulting in build errors in the continuous integration tasks for Linux.

spnngl:
Done

lelegard:
Hi @spnngl,

I have added option `srt` to `--version` in all tools. The optional value after `--option ` is not documented but it can be very useful sometimes.

See on my Mac:
~~~
$ tsp --version=srt
libsrt version 1.4.1
$
$ tsp --version=all
tsp: TSDuck - The MPEG Transport Stream Toolkit - version 3.20-1608
Built Jan 15 2020 - 22:37:35
Using Clang 11.0.0 (clang-1100.0.33.17), C++ std 2011.03
Web library: libcurl: 7.64.1, ssl: SecureTransport (LibreSSL/2.8.3), libz: 1.2.11
SRT library: libsrt version 1.4.1
Dektec: This version of TSDuck was compiled without Dektec support
$
$ tsp --version=x
tsp: invalid value x for option --version, use one of "all", "compiler", "date", "dektec", "http", "integer", "long", "nsis", "short", "srt"
$
~~~

lelegard:
Hi @spnngl 

I would like to add an SRT example in the TSDuck user's guide. And I also would like to see it working. But I am ignorant about SRT.

Could you please give me appropriate commands to test and illustrate the usage of SRT?

Typically, this would be something like this. From sender system:
~~~
tsp -I file input.ts --infinite -P regulate -O srt ?????
~~~
And from receiver system:
~~~
tsp -I srt ????? -P until --seconds 30 -P analyze -O drop
~~~

I just need actual examples for the `?????`.
Thanks

spnngl:
Hi, thanks for the doc ! :)

The simplest commands would be :
`tsp -I file input.ts --infinite -P regulate -O srt 0.0.0.0:4242 --transtype live --messageapi`
`tsp -I srt 192.168.10.10:4242 --transtype live --messageapi -P until --seconds 30 -P analyze -O drop`

Another example with encrypted transmission :
`tsp -I file input.ts --infinite -P regulate -O srt 0.0.0.0:4242 --transtype live --messageapi --passphrase baguettes --pbkeylen 16 --enforce-encryption --min-version 0x010400`
`tsp -I srt 192.168.10.10:4242 --transtype live --messageapi --passphrase baguettes --pbkeylen 16 -P until --seconds 30 -P analyze -O drop`

lelegard:
Thanks, I will try them between two Linux VM, or between the Mac host and one VM.

Just one question, due to my ignorance of SRT, when you write _"the listener, IP 192.168.10.10, listens on its port 4242"_, is the listener the one which _outputs_ with `-O srt 0.0.0.0:4242`?

My understanding is that the sender listens for an incoming connection and then sends the stream over it. This is a bit counter-intuitive compared to IP TV (multicast) where the listener is the receiver.

spnngl:
Yes, it is exactly that

cbernander:
To clarify, in SRT, the concepts of caller and listener are separate from sender and receiver, i.e. the side that establishes the network connection (caller) to the other side (listener) can act as either sender or receiver once the connection has been established.

lelegard:
@cbernander, thanks for the clarification.

@spnngl, I have tested your examples between the macOS host and a Fedora Linux VM. It works but I just had to enlarge the passphrase to 10 characters, the minimum length.

Moreover, in commit 8417380a46e8232dcc3ccf225ded2dbff4c25f80, I fixed a few issues with srt parameters. First, three more options have been made boolean (options without values) instead of values 0 or 1. Worse, three strings parameters which are directly used as socket parameters (including the passphrase) were passed to libsrt as UTF-16 (UString implementation). The binary representation was both incorrect (with a zero every 2 characters) and a truncated size (the method size() returns the number of 16-bit characters, not the number of bytes). Concerning the passphrase, this was interoperable between two instances of TSDuck (bug-compatible) but not with other systems.

The documentation (examples and option fixes) has been updated in the next commit.

lelegard:
@spnngl, I also moved all dependencies to libsrt inside the implementation of the class `SRTSocket`. The class `SRTSocket` becomes available to all applications. But if the class was compiled without SRT support (no libsrt), an error message is reported at run time. Thus, applications and plugins do not have to worry about `TS_NOSRT`, this is hidden inside one class. This is a pattern I use often in TSDuck. This simplifies the applications and reduces the risk of compilation failure on systems with a unusual combination of installations.

lelegard:
Hi,

I have rebuilt TSDuck with SRT on Windows (I will let you know how I automated the shitty build of SRT on Windows). I tried the same sample commands between the macOS host and a Windows VM. Roughly, it works because the receiver can analyse the stream and retrieve its structure and services. However, I get a lot of error messages that I did not have between macOS and Linux.

Receiver on Windows: many messages with this pattern:
~~~
23:57:38.794000*E:SRT.c: %322283115:No room to store incoming packet: offset=9515 avail=8185 ack.seq=748984983 pkt.seq=748994498 rcv-remain=6
~~~

Sender on macOS: many messages with this pattern:
~~~
23:57:38.762106*E:SRT.d: SND-DROPPED 55 packets - lost delaying for 1030ms
~~~

I do not know what is the reason. I do not want to investigate too much. First reason is that I had to invest a lot of time to build libsrt on Windows and automate it. Second reason is that I am still very ignorant on SRT. So, I did my part of the job in making it built. I will let SRT users investigate the issue and fix it.

lelegard:
Let's continue the investigation of the problem in issue #506.

spnngl:
Thanks for all you work on SRT plugin ! Sorry for not responding a lot, I had not access to a computer this weekend

lelegard:
You're welcome. Do you known anyone who could investigate the Windows issue #506 ?

spnngl:
No sorry, only linux users around me

lelegard:
Hi @spnngl 

This may be just a detail you do not care about, but your contributions are not accounted to your account @spnngl (see https://github.com/tsduck/tsduck/graphs/contributors).

Another contributor had this problem once. The problem was that the author email in the commits did not match any registered email in GitHub. So, if you would like your contributions to be correctly credited to your account, you should add your commit email in the list of your registered email addresses in your GitHb account.

Your choice...

spnngl:
Thanks, I'll look into it. I never look at those stats myself ^^

PR #508, opened 2020-01-20
Update sample-app

spnngl:


lelegard:
Thanks. This sample application has been written a long time ago and never updated.

PR #510, opened 2020-01-22
SRT: Add rendez-vous mode

spnngl:
Done rendez-vous mode TODO.

### DIFF
--- a/src/libtsduck/base/network/tsSRTSocket.cpp
+++ b/src/libtsduck/base/network/tsSRTSocket.cpp
@@ -496,7 +496,7 @@ bool ts::SRTSocket::open(const ts::SocketAddress& local_addr,
             if (ret < 0) {
                 goto fail;
             }
-            /* FALLTHROUGH */
+            TS_FALLTHROUGH
         case CALLER:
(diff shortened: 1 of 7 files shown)


lelegard:
Thanks.

Could you please give a sample of sender and receiver commands in rendezvous mode? Same as you did for the initial release of the plugin.

What is the IP address parameter in that case?
- Is this the address of the peer (the sender specifies the receiver address and vice versa)?
- Or is this the address of a third system using some specific server to act as a proxy between the peers?

spnngl:
Rendez-vous mode is used by the two sides to connect at each other at the same time. Ports numbers have to be the same. 

For SRTInputPlugin the default parameter is the remote IP and the rendezvous IP is the local one. 

tsp -I srt 10.1.1.140:4242 —transtype live —messageapi —rendezvous 4242 -O drop

For SRTOutputPlugin the default parameter is the local IP and the rendezvous IP is the remote one. 

tsp -I<something> -P ... -O srt 4242 —transtype live —messageapi —rendezvous 10.1.1.130:4242

Sorry for the formatting, I will not have access to a computer until Monday

Issue #512, opened 2020-01-24
Last build fails with gcc-4.9


lars18th:
Hi @lelegard ,

Current repo fails to compile with the GCC-4.9 backport:
```
make CC=/usr/lib/gcc-4.9-backport/bin/gcc GCC=/usr/lib/gcc-4.9-backport/bin/gcc CXX=/usr/lib/gcc-4.9-backport/bin/g++ AR=/usr/lib/gcc-4.9-backport/bin/gcc-ar ARFLAGS=rc NOTEST=1 NOPCSC=1 NODTAPI=1

[...]
  [CXX] dtv/descriptors/tsAbstractDeliverySystemDescriptor.cpp
In file included from /home/user/tsduck/src/libtsduck/dtv/tsModulation.h:37:0,
                 from /home/user/tsduck/src/libtsduck/dtv/tsModulationArgs.h:39,
                 from dtv/descriptors/tsAbstractDeliverySystemDescriptor.h:37,
                 from dtv/descriptors/tsAbstractDeliverySystemDescriptor.cpp:30:
/home/user/tsduck/src/libtsduck/dtv/tsDeliverySystem.h:53:28: error: '::SYS_DVBC_ANNEX_A' has not been declared
         DS_DVB_C_ANNEX_A = ::SYS_DVBC_ANNEX_A,
                            ^
/home/user/tsduck/src/libtsduck/dtv/tsDeliverySystem.h:55:28: error: '::SYS_DVBC_ANNEX_C' has not been declared
         DS_DVB_C_ANNEX_C = ::SYS_DVBC_ANNEX_C,
                            ^
/home/user/tsduck/src/libtsduck/dtv/tsDeliverySystem.h:63:28: error: '::SYS_DTMB' has not been declared
         DS_DTMB          = ::SYS_DTMB,
                            ^
In file included from /home/user/tsduck/src/libtsduck/dtv/tsModulationArgs.h:39:0,
                 from dtv/descriptors/tsAbstractDeliverySystemDescriptor.h:37,
                 from dtv/descriptors/tsAbstractDeliverySystemDescriptor.cpp:30:
/home/user/tsduck/src/libtsduck/dtv/tsModulation.h:111:20: error: '::QAM_4_NR' has not been declared
         QAM_4_NR = ::QAM_4_NR,
                    ^
/home/user/tsduck/src/libtsduck/dtv/tsModulation.h:200:20: error: '::FEC_2_5' has not been declared
[...]

PR #513, opened 2020-01-24
Kernel legacy compatibility 

lars18th:
Restore compatbility with some legacy kernels (linux 3.2) because some missing declarations.

### DIFF
--- a/src/libtsduck/base/tsPlatform.h
+++ b/src/libtsduck/base/tsPlatform.h
@@ -858,6 +858,7 @@ TS_MSC_NOWARNING(5027)
 #include <linux/dvb/version.h>
 #include <linux/dvb/frontend.h>
 #include <linux/dvb/dmx.h>
+#include <linux/version.h>
 #endif
 
 #if defined(TS_MAC)
--- a/src/libtsduck/dtv/linux/tsDTVProperties.h
+++ b/src/libtsduck/dtv/linux/tsDTVProperties.h
@@ -36,6 +36,10 @@
 #include "tsReport.h"
 #include "tsVariable.h"
 
+#if !defined(NO_STREAM_ID_FILTER)
+#define NO_STREAM_ID_FILTER    (~0U)
+#endif
+
 namespace ts {
     //!
     //! Encapsulation of Linux DVB property lists.
--- a/src/libtsduck/dtv/linux/tsTunerGuts.cpp
+++ b/src/libtsduck/dtv/linux/tsTunerGuts.cpp
@@ -540,7 +540,9 @@ bool ts::Tuner::Guts::getCurrentTuning(ModulationArgs& params, bool reset_unknow
             props.add(DTV_MODULATION);
             props.add(DTV_PILOT);
             props.add(DTV_ROLLOFF);
+#if defined(DTV_STREAM_ID)
             props.add(DTV_STREAM_ID);
+#endif
 
             if (::ioctl(frontend_fd, ioctl_request_t(FE_GET_PROPERTY), props.getIoctlParam()) < 0) {
                 const ErrorCode err = LastErrorCode();
@@ -556,7 +558,11 @@ bool ts::Tuner::Guts::getCurrentTuning(ModulationArgs& params, bool reset_unknow
             params.roll_off = RollOff(props.getByCommand(DTV_ROLLOFF));
 
             // With the Linux DVB API, all multistream selection info are passed in the "stream id".
+#if defined(DTV_STREAM_ID)
             const uint32_t id = props.getByCommand(DTV_STREAM_ID);
+#else
+            const uint32_t id = NO_STREAM_ID_FILTER;
+#endif
             params.isi = id & 0x000000FF;
             params.pls_code = (id >> 8) & 0x0003FFFF;
             params.pls_mode = PLSMode(id >> 26);
@@ -574,7 +580,9 @@ bool ts::Tuner::Guts::getCurrentTuning(ModulationArgs& params, bool reset_unknow
             props.add(DTV_TRANSMISSION_MODE);
             props.add(DTV_GUARD_INTERVAL);
             props.add(DTV_HIERARCHY);
+#if defined(DTV_STREAM_ID)
             props.add(DTV_STREAM_ID);
+#endif
 
             if (::ioctl(frontend_fd, ioctl_request_t(FE_GET_PROPERTY), props.getIoctlParam()) < 0) {
                 const ErrorCode err = LastErrorCode();
@@ -591,7 +599,11 @@ bool ts::Tuner::Guts::getCurrentTuning(ModulationArgs& params, bool reset_unknow
             params.transmission_mode = TransmissionMode(props.getByCommand(DTV_TRANSMISSION_MODE));
             params.guard_interval = GuardInterval(props.getByCommand(DTV_GUARD_INTERVAL));
             params.hierarchy = Hierarchy(props.getByCommand(DTV_HIERARCHY));
+#if defined(DTV_STREAM_ID)
             params.plp = props.getByCommand(DTV_STREAM_ID);
+#else
+            params.plp = NO_STREAM_ID_FILTER;
+#endif
             return true;
         }
         case DS_DVB_C_ANNEX_A:
@@ -1013,7 +1025,9 @@ bool ts::Tuner::tune(ModulationArgs& params, Report& report)
                     ((params.pls_code.value(ModulationArgs::DEFAULT_PLS_CODE) & 0x0003FFFF) << 8) |
                     (params.isi.value() & 0x000000FF);
                 report.debug(u"using DVB-S2 multi-stream id 0x%X (%d)", {id, id});
+#if defined(DTV_STREAM_ID)
                 props.add(DTV_STREAM_ID, id);
+#endif
             }
             break;
         }
@@ -1029,7 +1043,9 @@ bool ts::Tuner::tune(ModulationArgs& params, Report& report)
             props.addVar(DTV_TRANSMISSION_MODE, params.transmission_mode);
             props.addVar(DTV_GUARD_INTERVAL, params.guard_interval);
             props.addVar(DTV_HIERARCHY, params.hierarchy);
+#if defined(DTV_STREAM_ID)
             props.addVar(DTV_STREAM_ID, params.plp);
+#endif
             break;
         }
         case DS_DVB_C_ANNEX_A:
--- a/src/libtsduck/dtv/linux/tsTunerGuts.cpp
+++ b/src/libtsduck/dtv/linux/tsTunerGuts.cpp
@@ -324,11 +324,15 @@ bool ts::Tuner::open(const UString& device_name, bool info_only, Report& report)
 
     _delivery_systems.clear();
     DTVProperties props;
+#if defined(DTV_ENUM_DELSYS)
     props.add(DTV_ENUM_DELSYS);
     if (::ioctl(_guts->frontend_fd, ioctl_request_t(FE_GET_PROPERTY), props.getIoctlParam()) >= 0) {
         // DTV_ENUM_DELSYS succeeded, get all delivery systems.
         props.getValuesByCommand(_delivery_systems, DTV_ENUM_DELSYS);
     }
+#else
+    if (0) {;}
+#endif
     else {
         // DTV_ENUM_DELSYS failed, convert tuner type from FE_GET_INFO.
         const ErrorCode err = LastErrorCode();
--- a/src/libtsduck/dtv/linux/tsTunerGuts.cpp
+++ b/src/libtsduck/dtv/linux/tsTunerGuts.cpp
@@ -1581,7 +1581,11 @@ std::ostream& ts::Tuner::displayStatus(std::ostream& strm, const ts::UString& ma
         {u"8-VSB",                  ::FE_CAN_8VSB},
         {u"16-VSB",                 ::FE_CAN_16VSB},
         {u"extended caps",          ::FE_HAS_EXTENDED_CAPS},
+#if LINUX_VERSION_CODE < KERNEL_VERSION(3,7,0)
+        {u"multistream",            int(0x4000000)},
+#else
         {u"multistream",            ::FE_CAN_MULTISTREAM},
+#endif
         {u"turbo FEC",              ::FE_CAN_TURBO_FEC},
         {u"2nd generation",         ::FE_CAN_2G_MODULATION},
         {u"needs bending",          ::FE_NEEDS_BENDING},
--- a/src/libtsduck/dtv/tsDeliverySystem.h
+++ b/src/libtsduck/dtv/tsDeliverySystem.h
@@ -50,8 +50,14 @@ namespace ts {
         DS_DVB_S_TURBO   = ::SYS_TURBO,
         DS_DVB_T         = ::SYS_DVBT,
         DS_DVB_T2        = ::SYS_DVBT2,
+#if !defined(SYS_DVBC_ANNEX_A)
+#define SYS_DVBC_ANNEX_A     SYS_DVBC_ANNEX_AC
+#endif
         DS_DVB_C_ANNEX_A = ::SYS_DVBC_ANNEX_A,
         DS_DVB_C_ANNEX_B = ::SYS_DVBC_ANNEX_B,
+#if !defined(SYS_DVBC_ANNEX_C)
+#define SYS_DVBC_ANNEX_C     SYS_DVBC_ANNEX_AC
+#endif
         DS_DVB_C_ANNEX_C = ::SYS_DVBC_ANNEX_C,
         DS_DVB_C2        = -10,
         DS_DVB_H         = ::SYS_DVBH,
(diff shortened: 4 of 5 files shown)


lars18th:
Hi @lelegard ,

This patch fixes the compilation problem described in #512 . I hope you want to merge it. It's very easy to maintain the compatibility in this case... as only some missing names are required to redefine. Futhermore I think it's good to continue maintaining kernels using GCC version 4.9, so please don't discard this PR.

Regards.

lelegard:
Sorry, but this is incorrect. You test conditions such as `#if !defined(SYS_DVBC_ANNEX_A)`. But `SYS_DVBC_ANNEX_A` is never defined as a preprocessor macro, it is an enum value. So, since it is never defined as a preprocessor macro, the code is never compiled, even on recent kernels. So the code is completely broken.

The right test should be to test the DVB API version (there are definitions for that, see [here](https://github.com/tsduck/tsduck/blob/master/src/libtsduck/base/tsPlatform.h#L994)). But, in that case, for each enum, you need to know the exact DVB API version at which it appeared and test each value against its own DVB API version.

> Futhermore I think it's good to continue maintaining kernels using GCC version 4.9, 

Again, as I already mentioned in #512, this is not a matter of gcc version, this is matter of kernel version. TSDuck compiles correctly with gcc 4.8.5 which is older but with kernel 3.10 (CentOS 7.7).

Anyway, even Debian, a large organization with many volunteers, no longer supports Debian 7. As unique developer of TSDuck, I will not clutter the code with so many conditional compilations to support an obsolete system. You have to realise that, in the past, there was a lot of conditional compilation to support DVB API version 4, which was completely different from version 5 (so-called "S2-API"). The code was completely cluttered. When it became obvious that pre-V5 API had disappeared from all supported systems, I cleaned up the code and it was a real relief.

TSDuck started in 2005. The Linux and Windows DVB API have considerably evolved during 15 years (and they will probably continue to evolve). If I had carefully maintained all `#ifdef` from the beginning, the code would have become an unmaintainable plate of spaghetti. So, as years go by, I cleanup the code, removing support for obsolete systems, keeping only a sliding window of 5 or 6 years of support. This is required to keep a clean code base.

You should drop Debian 7 and upgrade to something more recent. This is not only a matter of TSDuck. If you have any problem with that system, even Debian will not support you. Please do not ask me to do what even Debian won't do.

lars18th:
Hi @lelegard ,

Let me to expose three relevant points:

1. Debian 7 uses DVB-API 5.0 as it includes the the kernel 3.2, and this is only 1 subversion less than the version _currently_ supported to compile TSDuck (5.1 and up). And this is not the first time I do some small changes in open source projects to include _explicit_ support for DVB-API 5.0, as it's very common to assume the 5.1 version as the standard. So from my point of view it's positive to test and compile in Debian 7. Sorry but I don't aggree with the sentence: _"You should drop Debian 7 and upgrade to something more recent"_. I use this platform because is DVB-API 5.0, and it's very good to test compatibility with DVB-API 5.0.
2. Futhermore, I'm not suggesting to add a lot of conditional code to support DVB-API 5.0. I'm only suggesting a patch for **7** _missing_ modulations not defined in DVB-API 5.0, and to add support for **3** delivery systems defined _differently_.
3. Regarding the delivery systems, please check the code of the stock Linux kernel 4.17:
   - `/* backward compatibility definitions for delivery systems */`
      `#define SYS_DVBC_ANNEX_AC       SYS_DVBC_ANNEX_A`
      `#define SYS_DMBTH               SYS_DTMB /* DMB-TH is legacy name, use DTMB */`
   From: https://www.kernel.org/doc/html/v4.17/media/uapi/dvb/headers.html.
   - So, in order to enable compatibility with software using DVB-API 5.0 with recent kernels two new defines are included: `SYS_DVBC_ANNEX_AC`and `SYS_DMBTH`. So targeting this special case in my patch I'm only adding a workaround based on this concept.

After I say this, I suggest you to reconsider the suggestion to incorporate a small change to _restore_ the support for DVB-API 5.0. From my point of view it's a very small effort. Futhermore if you're not satisfied with my current proposal, we can do changes. Remember: two different concepts are targeted: the legacy delivery systems definition, and the new modulations not incorporated. For the first case, please comment your suggestion. And for the second case, I think my proposal for the code is quite clean.

And finally, take note that the continous integration has failed to check my patch with Linux because the _"Install dependencies"_ has failed (in fact, because you need to remove the `apt-get update`), and not because my fault.

Please, check it.
Regards.

lelegard:
Anyway, the PR with checking the macro-definition of enum values is incorrect. You have to browse the kernel history between 3.2 and 3.10. For each enum value, check the precise version of the DVB API for which it appears. Then, condition each enum value for its initial DVB API value.

This is a lot of work for an obsolete kernel version.

Again, I see no reason why I should do a piece of work that even a large and renown community such as Debian does not want to do. Even a commercial company such as Red Hat, with _paying_ customers is very restrictive about obsolete kernels: RH 8.x uses only kernel 4.18, RH 7.x only uses 3.10. The only supported RH version with older kernels is RH 6.10 with kernel 2.6.32 (and only up to 2021). Nobody, even RH with its _paying_ customers, supports kernel 3.2. With RH, it's only 2.6.32 (up to 2021), 3.10 or 4.18, no intermediate version.

With Debian 7, you are all alone, on your own. Debian does not support you. Your problem is not compiling TSDuck on Debian 7. Your problem is the trouble of using a distro with noone supporting you, neither Debian nor me.

PS: Do not worry about CI failures, all users of GitHub Actions seem to have run into this during a limited period of time that day. I have seen this in the GitHub Actions forum.

lars18th:
Hi @lelegard ,

Of course I don't want to get into a heated discussion. Please, note that what I'm commenting is _RESTORE THE SUPPORT FOR DVB-API 5.0_. Nothing more.

The use of Debian 7 is my preference (not related to anything else), but the fact is that it's a good testing platform of **DVB API 5.0**. So my objective is only this one.

Regarding the _errors_ in my patch, I'll fix them. However, I ask you before if you agree to incorporate these _small_ changes to continue supporting DVB-API 5.0. We don't need to check every kernel version/subversion and the DVB-API version... as the _differences_ between them are frozen. So simple defines can overcome the problem, and it's only required to check the compilation in _one_ platform to ensure the compatibility.

Please let me know if you accept this or not.
In any case, thank you for this awesome project. As you're the main develop I'll accept your decision.
Regards.

lars18th:
Hi @lelegard ,

As usual, you're right, and I'm wrong!
So I used your suggestions and I updated the patch.

Please note:
- The problem is not related to the version of DVB-API. It's only a problem with new declarations included in recent kernels. So, this patch _restores compatibility with kernels starting from 3.0_. Without this patch the mininal required version is Linux 3.7.0. Just as a simple remark: 3.16.81 is the older LTS kernel with support at time.
- All checks are done with recent kernels (using the CI) and with Debian 7 x64 using 3.2.96. Now the patch is clean.
- I'm only not fully satisfied with the solution for old kernels when declaring `DS_DVB_C_ANNEX_C`. You agree with my proposal?
- You can use the base of this patch to support _recent_ kernels with new functionalities in the future. I feel the code is clean and easy to maintain. You agree?

I hope you want to support kernels from 3.0. I understand you're only one developer in spare time. And for this specific reason I want to contribute.
Please, comment it.
Regards.

lars18th:
Hi @lelegard ,

Great! And good job with https://github.com/tsduck/tsduck/commit/480fd69ee19348d736c7b363d46a20695d041d8e _(the `if(0)` is because my C and not C++ background)_.

However, one question: _"I'm only not fully satisfied with the solution for old kernels when declaring `DS_DVB_C_ANNEX_C`."_ Do you feel that my solution is correct?

lars18th:
Hi @lelegard ,

Regarding my previos message, I found one error regarding this special case:
```
make test
source release-x86_64/setenv.sh && release-x86_64/utest
: invalid value DVB-C for option --delivery-system, use one of "ATSC", "ATSC-MH", "CMMB", "DAB", "DSS", "DTMB", "DVB-C", "DVB-C/A", "DVB-C/B", "DVB-C/C", "DVB-C2", "DVB-H", "DVB-S", "DVB-S-Turbo", "DVB-S2", "DVB-T", "DVB-T2", "ISDB-C", "ISDB-S", "ISDB-T", "undefined"
make[1]: *** [test] Error 1
make: *** [test] Error 2
```

So, I feel I can't do the simple
```
#if LINUX_VERSION_CODE < KERNEL_VERSION(3,3,0)
        DS_DVB_C_ANNEX_A = ::SYS_DVBC_ANNEX_AC,
#else
        DS_DVB_C_ANNEX_A = ::SYS_DVBC_ANNEX_A,
#endif
        DS_DVB_C_ANNEX_B = ::SYS_DVBC_ANNEX_B,
#if LINUX_VERSION_CODE < KERNEL_VERSION(3,3,0)
        DS_DVB_C_ANNEX_C = -11,
#else
        DS_DVB_C_ANNEX_C = ::SYS_DVBC_ANNEX_C,
#endif
```
But I do it in this way as if I repeat the `SYS_DVBC_ANNEX_AC` then the compilation fails.

Any idea to fix this problem?

lars18th:
Hi @lelegard ,

If I interchange the code (`DS_DVB_C_ANNEX_A = -11` and `DS_DVB_C_ANNEX_C = ::SYS_DVBC_ANNEX_AC`) then executing all tests in Debian 7 with:
```
 make CC=/usr/lib/gcc-4.9-backport/bin/gcc GCC=/usr/lib/gcc-4.9-backport/bin/gcc CXX=/usr/lib/gcc-4.9-backport/bin/g++ AR=/usr/lib/gcc-4.9-backport/bin/gcc-ar ARFLAGS=rc test-suite
[...]
----  ALL 884 tests passed
```

So, we only need to fix this remaining trouble.
However, I don't have the solution.
Can you help please?

lelegard:
What is the problem with the current code (master repo):
~~~
#if LINUX_VERSION_CODE < KERNEL_VERSION(3,3,0)
        DS_DVB_C_ANNEX_A = ::SYS_DVBC_ANNEX_AC,
#else
        DS_DVB_C_ANNEX_A = ::SYS_DVBC_ANNEX_A,
#endif
        DS_DVB_C_ANNEX_B = ::SYS_DVBC_ANNEX_B,
#if LINUX_VERSION_CODE < KERNEL_VERSION(3,3,0)
        DS_DVB_C_ANNEX_C = -11,
#else
        DS_DVB_C_ANNEX_C = ::SYS_DVBC_ANNEX_C,
#endif
~~~
Before kernel 3.3, `SYS_DVBC_ANNEX_AC` is an enum value. With 3.3 onward, it is a macro definition pointing to `SYS_DVBC_ANNEX_A`. So, the code should be ok.

lelegard:
As a reference, you may check the attached file: [dvbapi-headers-history.zip](https://github.com/tsduck/tsduck/files/4143358/dvbapi-headers-history.zip)

The shell script does the following:
- Clone the master kernel repo from `kernel.org`.
- Checkout all main X.X.X versions of the kernel, one by one, and extract the DVB API header files.
- Remove all duplicates, ie. consecutive versions of the kernel with identical DVB API headers.

The result is only 42 directories with 8 header files each (they are all in the zip file). It is then very easy to track the evolution of the various symbols:
~~~
$ grep ANNEX_AC kernel_*/*
kernel_2_06_028_dvbapi_5_00/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_2_06_029_dvbapi_5_00/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_2_06_030_dvbapi_5_00/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_2_06_032_dvbapi_5_01/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_2_06_036_dvbapi_5_02/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_3_00_000_dvbapi_5_03/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_3_01_000_dvbapi_5_03/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_3_02_000_dvbapi_5_04/frontend.h:	SYS_DVBC_ANNEX_AC,
kernel_3_03_000_dvbapi_5_05/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_05_000_dvbapi_5_06/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_000_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_001_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_002_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_003_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_004_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_005_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_006_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_007_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_008_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_009_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_07_010_dvbapi_5_09/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_09_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_3_10_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_02_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_05_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_09_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_12_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_13_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_14_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_15_000_dvbapi_5_10/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_16_000_dvbapi_5_11/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_4_19_000_dvbapi_5_11/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
kernel_5_03_000_dvbapi_5_11/frontend.h:#define SYS_DVBC_ANNEX_AC	SYS_DVBC_ANNEX_A
~~~

lars18th:
> What is the problem with the current code (master repo):

The problem is quite obscure... Let me to explain it:

- With the current code, the TSDuck library compiles OK.
- However, when compiling and executing the _tests_ comething fails (test 4). See this message: `: invalid value DVB-C for option --delivery-system, use one of "ATSC", "ATSC-MH", "CMMB", "DAB", "DSS", "DTMB", "DVB-C", "DVB-C/A", "DVB-C/B", "DVB-C/C", "DVB-C2", "DVB-H", "DVB-S", "DVB-S-Turbo", "DVB-S2", "DVB-T", "DVB-T2", "ISDB-C", "ISDB-S", "ISDB-T", "undefined"`
- And if I try to use `#define DS_DVB_C_ANNEX_C   DS_DVB_C_ANNEX_A` then the compilation of the library fails because in the code at some points appears this:
` case DS_DVB_C_ANNEX_A:
  case DS_DVB_C_ANNEX_C:`
And this is valid code, but it can't be compiled because both identifiers share the same value.

So, in order to maintain clean code, DS_DVB_C_ANNEX_A & DS_DVB_C_ANNEX_C **can't share the same value**. But, when doing in this way some of the tests fails.

Then I ask you for a balancing solution. In case we need to break something I prefer to define `DS_DVB_C_ANNEX_A` with a negative value and `DS_DVB_C_ANNEX_C` as `SYS_DVBC_ANNEX_AC` because at time the tests compile with no errors. However, I don't know about other side effects with this solution.

What's your opinion?

lelegard:
> However, when compiling and executing the tests comething fails (test 4). 

Please be more specific. Which test 4? In which environment? Before or after kernel 3.3? Or both?

> And if I try to use #define DS_DVB_C_ANNEX_C DS_DVB_C_ANNEX_A

No, you can't do that, this is inconsistent. 

All these efforts and time for obsolete unsupported kernels. I knew I should have stayed to my original position: obsolete and no longer supported, full stop.

PR #524, opened 2020-02-07
Parse all SEI even if there are more than one in a NALU

MartinFretigne:
Only the first SEI was parsed, this PR would allow the pes filter to parse all of them.


lelegard:
Thanks for the fix. Sorry for the merge delay.

PR #525, opened 2020-02-07
Parse and display SEI pic_timing

MartinFretigne:
This PR adds code to parse and display the field of AVC SEI pic_timing. I'm not sure if its the right location for that code, but that's a feature I would lke tsduck to contain.


MartinFretigne:
Closing because the ts::AVCParser should probably be used instead of ts::BitStream, to properly handle emulation_prevention_three_byte.

PR #526, opened 2020-02-07
Inject time signal at specific time using splice time and start delay

MartinFretigne:
The aim of this PR is to inject time_signal 4s before pts_time instead of right at the beginning.
The spliceinject filter options used are : -P spliceinject --service 1545  --inject-count 1 --files '/tmp/splices000.xml' --wait-first-batch

splices000.xml contained this : 
```
 <splice_information_table>
 <time_signal pts_time="6957864362" />
 <splice_segmentation_descriptor segmentation_event_id="100" segmentation_type_id="0x24" segment_num="1" segments_expected="1">
 <segmentation_upid type="0"/>
 </splice_segmentation_descriptor>
 </splice_information_table>
```

### DIFF
--- a/src/tsplugins/tsplugin_spliceinject.cpp
+++ b/src/tsplugins/tsplugin_spliceinject.cpp
@@ -799,6 +799,9 @@ ts::SpliceInjectPlugin::SpliceCommand::SpliceCommand(SpliceInjectPlugin* plugin,
             next_pts = (last_pts - preceding) & PTS_DTS_MASK;
         }
     }
+    else if (sit.isValid() && sit.splice_command_type == SPLICE_TIME_SIGNAL) {
+       next_pts = sit.time_signal.value() - 90000*4;
+    }
 }
 
 
--- a/src/tsplugins/tsplugin_spliceinject.cpp
+++ b/src/tsplugins/tsplugin_spliceinject.cpp
@@ -769,23 +769,31 @@ ts::SpliceInjectPlugin::SpliceCommand::SpliceCommand(SpliceInjectPlugin* plugin,
 
     // The initial values for the member fields are set for one immediate injection.
     // This must be changed for non-immediate splice insert commands.
-    if (sit.isValid() && sit.splice_command_type == SPLICE_INSERT && !sit.splice_insert.canceled && !sit.splice_insert.immediate) {
+    if (sit.isValid() && ((sit.splice_command_type == SPLICE_TIME_SIGNAL) || 
+        (sit.splice_command_type == SPLICE_INSERT && !sit.splice_insert.canceled && !sit.splice_insert.immediate))) {
         // Compute the splice event PTS value. This will be the last time for
         // the splice command injection since the event is obsolete afterward.
-        if (sit.splice_insert.program_splice) {
-            // Common PTS value, program-wide.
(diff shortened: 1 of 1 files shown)


lelegard:
Hi @MartinFretigne 

I am trying to understand the rationale for this PR.

The description of the time_signal() command in the SCTE-35 specification is not very clear to me. I may be wrong but my understanding of the time_signal() command is that this provides an alternate source of PTS (instead of the PTS from the video PID). So, in that case, the time_signal() is not synchronised on the video PID, it is another source of synchronisation. It just must be set "on time", which makes it not a good candidate for the `spliceinject` plugin.

Your PR seems to suggest that I am wrong (I can accept that 😃) and that the time_signal() shall be synchronised with the PTS of the video PID. In that case, don't you think that the time_signal() should be processed the same way as a splice_insert()? 

So, the time_signal() should set `next_pts`, `last_pts` and `count`, just like a splice_insert(). But just inserting it once at a hard-coded offset of 4 seconds does not seem a good idea. Or is there anything like this in the standard that I missed?

One last point: in the following code:
~~~
    else if (sit.isValid() && sit.splice_command_type == SPLICE_TIME_SIGNAL) {
        next_pts = sit.time_signal.value() - 90000*4;
     }
~~~
Take care that `sit.time_signal` is a subclass of `Variable<uint64_t>`. This means that the value may be uninitialised (the equivalent of the bit `time_specified_flag` set to zero in the splice_time() structure). So, you must first check that the value is `set()` before calling `value()`.

MartinFretigne:
Hi Thierry, thank you for reviewing this PR. 
> The description of the time_signal() command in the SCTE-35 specification is not very clear to me. I may be wrong but my understanding of the time_signal() command is that this provides an alternate source of PTS (instead of the PTS from the video PID). So, in that case, the time_signal() is not synchronised on the video PID, it is another source of synchronisation. It just must be sent "on time", which makes it not a good candidate for the `spliceinject` plugin.
> 
> Your PR seems to suggest that I am wrong (I can accept that smiley) and that the time_signal() shall be synchronised with the PTS of the video PID. 
Indeed time_signal can be used to perform similar actions to splice_insert messages, such as triggering splicing and triggering IDR insertion by an encoder.

> In that case, don't you think that the time_signal() should be processed the same way as a splice_insert()?
> 
> So, the time_signal() should set `next_pts`, `last_pts` and `count`, just like a splice_insert(). But just inserting it once at a hard-coded offset of 4 seconds does not seem a good idea. Or is there anything like this in the standard that I missed?
Yes, I agree with you, I should have implemented that as you suggest. 
 
> One last point: in the following code:
> 
> ```
>     else if (sit.isValid() && sit.splice_command_type == SPLICE_TIME_SIGNAL) {
>         next_pts = sit.time_signal.value() - 90000*4;
>      }
> ```
> 
> Take care that `sit.time_signal` is a subclass of `Variable<uint64_t>`. This means that the value may be uninitialised (the equivalent of the bit `time_specified_flag` set to zero in the splice_time() structure). So, you must first check that the value is `set()` before calling `value()`.

Ok thanks, I will rework that patch.

MartinFretigne:
I just checked the result of the new version of this PR, the time signal message is inserted where it's supposed to (splice-time minus start-delay), so it's ready to be reviewed again !

Issue #527, opened 2020-02-11
rookie question: how to enable SRT


J-Rogmann:
Ahoi
I got a little rookie question on how to enable SRT when building tsduck from source. 
I'm on Ubuntu 18.04 and did **NOT** specify `NOSRT=1` as build option. But still it says `* Error: srt: This version of TSDuck was compiled without SRT support` when trying to use SRT protocol.
My steps were:
```
git clone https://github.com/tsduck/tsduck.git
build/install-prerequisites.sh
make NOPCSC=1 NOCURL=1 NODTAPI=1
source src/tstools/release-x86_64/setenv.sh
```
Are there any special instructions to enable SRT support prior to building? I noticed, for windows there is a special script to download a SRT library build called `install-libsrt.ps1 ` . Is there something equivalent to do on Linux?

best regards,
Justus

lelegard:
Hi @J-Rogmann,

Ubuntu starts to support SRT in its official repo with 19.04 (package `libsrt-dev`, version 1.4.1).

With Ubuntu 18.04, there is no SRT base support. If you cannot upgrade Ubuntu, then you have to rebuild and install libsrt yourself (see https://github.com/Haivision/srt).

Once you have manually installed libsrt, just restart `make`. Libsrt will be automatically discovered and used in TSDuck. Note: start with `make distclean` to ensure you restart the build from scratch.

lelegard:
>  I noticed, for windows there is a special script to download a SRT library build called install-libsrt.ps1 . Is there something equivalent to do on Linux?

On Linux and macOS, the script `install-prerequisites.sh` installs the appropriate libsrt package when there is one for the system on which it is executed (none for Ubuntu 18.04).

lelegard:
Last advice: use `make -j10` or any appropriate number of parallel jobs. A pure sequential `make` is awfully long........

lelegard:
Oops, @J-Rogmann, I did not realize you were part of the Haivision/srt team. I assume there is no need to tell you how to build libsrt 😉

J-Rogmann:
Well, as I said: I'm still a rookie when it's comes to software DEV :)
I simply forgot to perform a `make install` on the SRT lib. 
Still worth noting for anybody else walking down this route: I still had to set the correct path to the SRT library for TSduck:
```
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib/
```
But still I have issues:
```
[...]

Issue #522, opened 2020-02-01
Compilation error when creating a non-install version


lars18th:
Hi @lelegard ,

This is related to #511, but is different.

When using different TSDuck versions, I create a subdirectory in the main directory (`~/tsduck-ver-X.XX/bin`). And I copy all binaries into this directory. Then I can use the tools using the full path. And then I don't mix versions as every binary directory is inside a different tree.

However, when a _new_ subdirectory exists the compilation fails. I prepared some fix with the PR #521 . However, this doesn't fix all the problems. So, please can you check the building in Linux when the user creates some subdirectory? I find it very annoying that this simple task completely breaks up the compilation.

Thank you!

lars18th:
And another question (I prefer to not open a lot of issues with related things):

- Can you add support for using some custom LD_LIBRARY_PATH, or use the current one?

When using the files `./src/utest/release-x86_64/setenv.sh` and `./src/tstools/release-x86_64/setenv.sh` I need to manually edit them when I use the GCC-4.9 compiler. The reason is because the libraries are installed in a non-standard path (to not mix different compilers, i.e. `/usr/lib/gcc-4.9-backport/lib`).

So the idea is to support something like: `make CUSTOM_LD_LIBRARY_PATH=<custom-dir>`; or to add to these files something like:
```
export LD_LIBRARY_PATH="/<install-dir>/src/libtsduck/release-x86_64:$LD_LIBRARY_PATH"
```

What you think?

lelegard:
The `setenv.sh` file is just a trick to run applications directly from the build. This is typically a debug trick (make, run, change code, make, run, etc.) This trick is needed just because the library, the plugins, the applications and the utests have different source locations and different object locations.

If you have several coexisting versions, it is simpler to just copy all binary files in the same directory. If everything is in the same directory, there is no need for `setenv.sh`.

The Unix make-based build is too complicated, I think that instead of having more setenv, more environment variables, I should simplify by compiling into one single binary directory, outside the source tree.

And BTW, to copy everything in the same directory, the command is `make install SYSPREFIX=...`

lars18th:
> I should simplify by compiling into one single binary directory, outside the source tree.

Great!

> And BTW, to copy everything in the same directory, the command is `make install SYSPREFIX=...`

I know it, but in my environment this generates other side effects. So for this reason I use my personal script. Perhaps a new update will remove the problems.

[...]

PR #529, opened 2020-02-12
Use user's system-wide LD_LIBRARY_PATH

lars18th:
When the user has the LD_LIBRARY_PATH defined, use it in the `setenv.sh`. If not then the execution fails if some used library isn't in the standard path.

### DIFF
--- a/src/tstools/Makefile
+++ b/src/tstools/Makefile
@@ -54,7 +54,7 @@ endif
 
 $(OBJDIR)/setenv.sh: Makefile
 	echo '[[ ":$$PATH:" != *:$(realpath $(OBJDIR)):* ]] && export PATH="$(realpath $(OBJDIR)):$$PATH"' >$@
-	echo 'export LD_LIBRARY_PATH="$(realpath $(LIBTSDUCKDIR)/$(OBJDIR))"' >>$@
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @lelegard ,

As commented here https://github.com/tsduck/tsduck/issues/522#issuecomment-581153342 this patch solves the problem when using libraries in non-default locations.

Futhermore, other cases like https://github.com/tsduck/tsduck/issues/527#issuecomment-584659426 are fixed too.

I hope you want to merge it.
Regards.

lelegard:
Hi,

The problem with this PR is that `LD_LIBRARY_PATH` is augmented again and again with the same value when `setenv.sh` is invoked several times. The test in the definition of `PATH` prevents this. The same type of test should be applied on all search paths, component by component.

Since this would make the various Makefiles too complicated, I implemented a more generic and more robust solution in commit 7052750fa8411f728b5857b6a8ad57a3bf2c708d.

I close this PR without merge since an alternate solution has been implemented.

lars18th:
> I close this PR without merge since an alternate solution has been implemented.

Great! I prefer your solution.
Thank you @lelegard another time.

PR #566, opened 2020-04-25
Some small changes to get the latest code base to build on VS2015

alevinsn:


alevinsn:
Note that while the tsARIBCharset change builds, I'm not sure if it was done correctly.

lelegard:
Hi Aaron,
Thanks, integrated.
The change is ARIBCharset is ok. Mainly it adds / converts a 32-bit Unicode code points into a UString, with UString containing Java-like UTF-16 codes.

Issue #571, opened 2020-05-04
Add Arch Linux support


piotrserafin:
Add Arch Linux Support:

- [x] PR with changes in install-prerequisites.sh accepted.
- [x] Update [AUR:tsduck](https://aur.archlinux.org/packages/tsduck/) package by me.

Above will allow Arch users to install TSDuck automatically from AUR repo.

piotrserafin:
[PR#571](https://github.com/tsduck/tsduck/pull/572) merged to master

lelegard:
Hi @piotrserafin,

Thanks a lot for your effort to provide TSDuck in Arch Linux. I do not use this distro and do not test on it. So, this is good to have a contributor to provide support on Arch.

For your information, I regularly test on Fedora, Ubuntu, CentOS and Raspbian. I provide binaries on all of them for official releases. I also have an Alpine Linux VM because this distro has a non-standard libc which creates compilation issues from time to time.

Out of curiosity, what is the packaging process for Arch Linux? Does it require specific packaging such as with `rpm` and `dpkg`? In that case, would you like to include the corresponding scripts or configuration files in the main source tree, as currently done to produce `.rpm` and `.deb` packages?

I see that AUR is the "Arch User Repository". Is this different from the main source of packages for Arch? I see that the available version here is 3.15 from November 2018. It is indeed time to upgrade.

piotrserafin:
Hello @lelegard ,

Packaging system in Arch is called ABS (Arch Build System) and it is used for compiling code to installable packages typically with .pkg.tar.xz extension. Those packages are maintained by core arch developers and kept in main package repositories. There is also AUR which contains packages maintained by general community, this is the place for tsduck. Arch users use pacman tool (you can see it in PR I've created) to install and manage system packages. 

ABS uses mapkepkg tool which reads PKGBUILD file and generates package based on it's content.
For example PKGBUILD for dvbblast (which can be found AUR) looks following: 

```bash
pkgname=dvblast
pkgver=3.4
pkgrel=1
pkgdesc="A simple and powerful MPEG-2/TS demux and streaming application"
arch=('i686' 'x86_64')
url="https://www.videolan.org/projects/dvblast.html"
license=('GPL2')
depends=('glibc' 'libev')
makedepends=('bitstream')
source=("https://get.videolan.org/dvblast/$pkgver/dvblast-$pkgver.tar.bz2")
[...]

PR #572, opened 2020-05-04
[#571] Added Arch Linux support

piotrserafin:
- Added Arch Linux support to install-prerequisites.sh script

### DIFF
--- a/build/install-prerequisites.sh
+++ b/build/install-prerequisites.sh
@@ -39,6 +39,7 @@
 #  - Fedora
 #  - Red Hat
 #  - CentOS
+#  - Arch
 #  - Alpine Linux
 #
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


piotrserafin:
CI Failed due to:

source release-x86_64/setenv.sh && release-x86_64/utest_static 
* Error: download error, SSL connect error, OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to tsduck.io:443 

*** HLSTest::testMasterPlaylist, assertion failure, utestHLS.cpp, line 107
condition: pl.loadURL(u"https://tsduck.io/download/test/hls/img_bipbop_adv_example_ts/master.m3u8", true)

*** FAILURES (1 tests FAILED, 422 passed, 1 assertions FAILED, 9836 passed)

lelegard:
Hi @piotrserafin 

Thanks, I will merge this.

Do not worry about the CI failure, there are intermittent network failures on the GitHub Runner systems. Most of the time, it fails during dependencies installation (intensive network usage) but also sometimes during tests. Several people (including myself) have opened topics about this.

I do not know Arch Linux. Are you sure that the package gcc installs C++ support? And are you sure that installing curl and pcsclite also installs their development counterparts (header files and libraries)? These questions are just to check since all other distros have different packages for gcc vs. g++ and applications vs. development packages.

piotrserafin:
Hello @lelegard,

Answering your questions:

1. gcc package includes both gcc and g++ binaries and all the headers, documentation, etc.
2. Arch works differently. The -dev packages (for Debian, RedHat based OSes) usually contain header-files, examples, documentation and such, which are not needed to just running the program (or use a library as a dependency). They are left out to save space. ArchLinux usually just ships these files with the package itself. This costs a bit more disk space for the installation but reduces the number packages you have to manage. Typically for Arch packages contains all the sources and PKGBUILD file which tells package management system how to build binaries from sources, where to put them after build step. 
3. Just to be sure I have used install_prerequisities.sh script and successfully build tsduck 
on fresh installation of Arch (headless).

PR #591, opened 2020-06-05
ContinuityAnalyzer:  Change from using private to protected to make i…

alevinsn:
…t easier to subclass the ContinuityAnalyzer class

### DIFF
--- a/src/libtsduck/dtv/tsContinuityAnalyzer.h
+++ b/src/libtsduck/dtv/tsContinuityAnalyzer.h
@@ -202,9 +202,9 @@ namespace ts {
         //!
         static int MissingPackets(int cc1, int cc2);
 
-    private:
+    protected:
         // PID analysis state
-        class PIDState
+        class TSDUCKDLL PIDState
         {
         public:
             PIDState();            // Constructor
@@ -217,7 +217,7 @@ namespace ts {
         // A map of PID state, indexed by PID.
         typedef std::map<PID,PIDState> PIDStateMap;
 
-        // Private members.
+        // Protected members.
         Report*       _report;            // Where to report errors, never null.
         int           _severity;          // Severity level for error messages.
         bool          _display_errors;    // Display discontinuity errors.


lelegard:
Hi Aaron,

Which services do you need in your subclass? Could we refine the need?

I am always a bit reluctant to break the data encapsulation principle because it changes the contract. When a modification is made inside the class, we can safely evaluate if the impact is limited to the class itself. If all private fields are protected, changing any implementation detail can break any subclass.

Which information do you need? Since you added the `TSDUCKDLL` attribute to `PIDState`, I suspect you need it. But do you need to get a copy of the state of a given PID? In which case, have a protected accesser returning a copy of it would be better. Or do you need to hack everything in the private part? In which case, you may have a problem in case of modification inside the base class.

alevinsn:
> Which services do you need in your subclass? Could we refine the need?

What I'm doing is non-standard, and it wouldn't be appropriate to include the functionality that I need in TSDuck proper.  Basically, I'm doing something that may cause packets to be duplicated once in an mpegts stream that had no duplicates to begin with.  The code in ContinuityAnalyzer::feedPacketInternal() notices these duplicates, and per the standard, correctly notices that there is only one duplicate and returns true, while also incrementing dup_count.  But, when I say my code can duplicate packets, it duplicates the packet entirely, _even_ if the packet includes a PCR.  Regardless, my code is doing something with the stream that it shouldn't be doing, but I can only detect it after the fact.  To do this, I subclassed ContinuityAnalyzer and added my own implementation of feedPacket(), as follows:

bool EnhancedContinuityAnalyzer::feedPacket(const ts::TSPacket& pkt, bool& bIsDuplicate)
{
	ts::PID pid = pkt.getPID();

	ts::TSPacket lastPacket = ts::NullPacket;
	if (_pid_states.find(pid) != _pid_states.end())
		lastPacket = _pid_states[pid].last_pkt_in;

	bool bRet = ContinuityAnalyzer::feedPacket(pkt);
	if (!bRet)
		return false;

	if (_pid_states[pid].dup_count > 0)
	{
		if (pkt == lastPacket)
		{
			// only mark duplicate packets that are entirely identical--if, say, the PCR is different, this should be considered
			// a normal scenario and disregarded
			bIsDuplicate = true;
		}
		else
			bIsDuplicate = false;
	}
	else
		bIsDuplicate = false;

	return true;
}

This code does not belong in TSDuck's ContinuityAnalyzer class and is only relevant to my situation.

> I am always a bit reluctant to break the data encapsulation principle because it changes the contract. When a modification is made inside the class, we can safely evaluate if the impact is limited to the class itself. If all private fields are protected, changing any implementation detail can break any subclass.

I understand that, although it's hard to imagine circumstances under which it would be necessary to change the implementation details of the ContinuityAnalyzer class.  Also, there have been many changes made to TSDuck in the last 6 months that have required TSDuck developers to change their code anyway.

> Which information do you need? Since you added the `TSDUCKDLL` attribute to `PIDState`, I suspect you need it. But do you need to get a copy of the state of a given PID? In which case, have a protected accesser returning a copy of it would be better. Or do you need to hack everything in the private part? In which case, you may have a problem in case of modification inside the base class.

I added TSDUCKDLL to PIDState to get my code to build.   PIDState class members can't be used without exporting the class.  As you can see from my custom implementation of feedPacket(), I only need read access to certain information in the class to accomplish my objective.  What I've done could be accomplished by exposing PIDState in a read-only fashion and providing a way to get a copy of the PIDState class or some or all of its members.  But, I would think the approach I took of using protected instead of private is preferable to doing that.

lelegard:
From your code, I would say that what you need in the protected interface is:

- Definition of `PIDState` (with the `TSDUCKDLL` attribute).
- Checking if a `PIDState` exists for a given PID value.
- A read-only accessor to the `PIDState` for a given PID value.

Such a protected interface seems quite reasonable.
What else would you need?

alevinsn:
> From your code, I would say that what you need in the protected interface is:
> 
> * Definition of `PIDState` (with the `TSDUCKDLL` attribute).
> * Checking if a `PIDState` exists for a given PID value.
> * A read-only accessor to the `PIDState` for a given PID value.
> 
> Such a protected interface seems quite reasonable.
> What else would you need?

Yeah, that will work.  I can submit a new pull request that accomplishes this.

PR #597, opened 2020-06-11
Protected2 and copy constructor continuity analyzer

alevinsn:
Rather than expose PIDState, I opted to just expose the per-PID duplicate count and last packet.  There were already methods for exposing the first and last CCs.

The enablement of the copy constructor is needed for forward packet CC analysis without affecting the "main" ContinuityCounter object for a stream.

### DIFF
--- a/src/libtsduck/dtv/tsContinuityAnalyzer.cpp
+++ b/src/libtsduck/dtv/tsContinuityAnalyzer.cpp
@@ -134,7 +134,7 @@ bool ts::ContinuityAnalyzer::hasPID(PID pid) const
 
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


alevinsn:
Also, I believe that I did the doxygen documentation correctly (including the one use of @ref), but I didn't confirm it with certainty by building with doxygen.

lelegard:
Hi Aaron,

Thanks for the fix. I merged it.

I had to fix a GCC warning (see the CI log). I also added a `getLastPacket()` in addition to your `lastPacket()` because returning a large structure by value is usually inefficient in the absence of a dedicated "move constructor" and "move assignment".

PR #599, opened 2020-06-13
 Fix multi-tuner detection on Linux and support Android device naming

### LINKED ISSUE #596: Linux Multi-Tuner Detection
Hi Thierry,

for Linux, there is a naming scheme for addressing a specific physical tuner:

https://github.com/tsduck/tsduck/blob/f82d2f3d6097a89982e0eef347ed42e910a2f140/src/libtsduck/dtv/tsTuner.h#L63-L67

but tslsdvb does not actually detect all the available physical tuners, just the adapters:

https://github.com/tsduck/tsduck/blob/da30e11a5a3c54e375d1b44c590cfbaed23f49d4/src/libtsduck/dtv/linux/tsTunerGuts.cpp#L233-L247

..so you'll never know how many (and which kinds of) physical tuners are available.

Is that just a missing bit or intentional for some reason?

SS-FF:
Fixes #596


lelegard:
Hi,

There are a few things I don't understand in the code. Did you test it on Linux and Android?

The function `replace()` seems equivalent to `UString::substitute()`.

`ExpandWildcard()` is called several times in a row. Each call erases the previous content of the output string array. Did you mean `ExpandWildcardAndAppend()`?

I don't understand this:
~~~
// If not specified, use frontend index for demux
if (fcount < 3) {
    demux_nb = frontend_nb;
}

// If not specified, use frontend index for dvr    
if (fcount < 4) {
    dvr_nb = frontend_nb;
}
~~~
It seems that, on Linux, `/dev/dvb/adapter2` will end up with demux `/dev/dvb/adapter2/demux2`, which seems incorrect to me.

Could you explain exactly what is the DVB device naming on Android? I assume that this is in fact "Android TV" on set-top box, right?

I thought that Android relied on a Linux kernel. So, why is the DVB device naming different? Are the DVB device drivers modified?

SS-FF:
> There are a few things I don't understand in the code. Did you test it on Linux and Android?
> `ExpandWildcard()` is called several times in a row. Each call erases the previous content of the output string array. Did you mean `ExpandWildcardAndAppend()`?

Yes exactly, that's the only flaw in the PR. I hadn't tested that last change on Android yet. 
We have some kind of super-static binaries that work on all (at least most) Linux distros and even on Android without doing a separate compilation. That's why I've merged the code and removed the conditional compilation (that I had added for Android) just few days ago.

> The function replace() seems equivalent to UString::substitute().

Thanks, I thought you must have something suitable, but the naming is s a bit unexpected :-)

> 
> I don't understand this:
> 
> ```
> // If not specified, use frontend index for demux
> if (fcount < 3) {
>     demux_nb = frontend_nb;
> }
> 
> // If not specified, use frontend index for dvr    
> if (fcount < 4) {
>     dvr_nb = frontend_nb;
> }
> ```
> 
> It seems that, on Linux, `/dev/dvb/adapter2` will end up with demux `/dev/dvb/adapter2/demux2`, which seems incorrect to me.

No, that's not the case. The number behind 'adapter' will never be parsed and will always exist as part of the string only.

- When `device_name` is `/dev/dvb/adapter2`, the `frontend_nb` will be `0` and `demux_nb` and `dvr_nb` will just be set to the same.
  => No change here

- When `device_name` is `/dev/dvb/adapter2:1`, the `frontend_nb` will be `1` and `demux_nb` and `dvr_nb` will be set to `1` as well.
  => That's the important change: It's a much better assumption that demux and dvr will have the same number as the frontend than assuming them to be always `0`.

I'm not sure whether cases even exist where demux and dvr would have different numbers, but those are definitely rare and in such cases, it's still possible to specify `/dev/dvb/adapter2:1:8:8`.

**GetAllTuners:**

`GetAllTuners()` will now always return device names including the frontend number as one of these patterns:

- `/dev/dvb/adapterX:Y`
  or
- `/dev/dvbX:Y`

> Could you explain exactly what is the DVB device naming on Android? I assume that this is in fact "Android TV" on set-top box, right?

Correct, but in order to access devices natively, you need either a rooted device or approval by the device vendor.

> I thought that Android relied on a Linux kernel. So, why is the DVB device naming different? Are the DVB device drivers modified?

I can't say why, but it seems that both naming variants exist:

https://android.googlesource.com/platform/frameworks/base/+/master/services/core/java/com/android/server/tv/TvInputManagerService.java#108

lelegard:
> > It seems that, on Linux, /dev/dvb/adapter2 will end up with demux /dev/dvb/adapter2/demux2, which seems incorrect to me.
>
> No, that's not the case. The number behind 'adapter' will never be parsed and will always exist as part of the string only.

Yes, you are right. There was a confusion in my mind between adapter number (never identified as an integer) and fontend number. Sorry about that.

> > I thought that Android relied on a Linux kernel. So, why is the DVB device naming different? Are the DVB device drivers modified?
>
> I can't say why, but it seems that both naming variants exist:
>
> https://android.googlesource.com/platform/frameworks/base/+/master/services/core/java/com/android/server/tv/TvInputManagerService.java#108

Did you ever encounter the "flat" structure without adapter directories? Do you think it is Android-specific or could it be some legacy Linux structure that the Android code included for completeness? 

In #596, you show a directory tree of one adapter with 16 frontends under `adapter0`. Was it produced on Android? Is it a regular Android TV or an AOSP one?

About this one, there are 16 frontends (which makes sense for 16 tuners) but there are 23 demux, 23 dvr, 23 net and 23 ca. 23 ? Why 23 ?

lelegard:
> Did you ever encounter the "flat" structure without adapter directories? Do you think it is Android-specific or could it be some legacy Linux structure that the Android code included for completeness?

OK, I can now answer this one. This is apparently a very old Linux legacy, prior to kernel 2.4. I have found references to that change in [Dec 2004 here](https://linuxtv.org/mailinglists/linux-dvb/2004/12-2004/msg00337.html). I have found other references from 2005 about users being puzzled about the change from flat to hierarchical structure.

I have also found references to the reason: before some kernel version (which one?), a driver could not create device nodes and directories and a static flat structure was required. After that kernel version, drivers were able to create device nodes and directories and then came the structured DVB device nodes.

Personally, I started to use DVB tuners on Linux in 2005 and never encountered the flat structure. Since it was long before the advent of Android, I do not think any Android device will ever see the flat structure (if the references I found are correct).

So, unless there is some evidence of the use of the flat structure in the last 10 years, I think we can simply drop the flat structure.

On the other hand, the modification looking for all fontends in `GetAllTuners` is still necessary.

SS-FF:
> So, unless there is some evidence of the use of the flat structure in the last 10 years, I think we can simply drop the flat structure.

10 years is the Linux story, but not the Android story...
Actually I'm working with a recent device, (delivered with Android Pie) that exposes the flat structure, otherwise I would have never come to that.

SS-FF:
> In #596, you show a directory tree of one adapter with 16 frontends under `adapter0`. Was it produced on Android? Is it a regular Android TV or an AOSP one?

That wasn't Android at all. It was armv8 Linux. Here are the results after this PR's change:

[FBC_new2.txt](https://github.com/tsduck/tsduck/files/4786006/FBC_new2.txt)

When you look at the device's product page that I referenced, it should become clear that it's not intended to be used like regular tuner hardware. It's some kind of set-top box that comes with its own software and not meant to run any other application. One could also say that: "we ran tslsdvb inside the firmware of the STB".

> About this one, there are 16 frontends (which makes sense for 16 tuners) but there are 23 demux, 23 dvr, 23 net and 23 ca. 23 ? Why 23 ?

I could explain 24, but not 23...

The device includes 2 PCIex modules: 
- 1 module providing 8 DVB-S tuners
- 1 module providing 8 tuners for either DVB-T or DVB-C

3x8 = 24... but not 23

Eventually it's probably irrelevant as it's not the kind of device meant to be accessed via DVB API.

lelegard:
> 10 years is the Linux story, but not the Android story...
Actually I'm working with a recent device, (delivered with Android Pie) that exposes the flat structure, otherwise I would have never come to that.

OK, weird. This is the first time I see the flat structure. So, just keep support for the flat structure. Since the driver decides the device structure, it is possible that drivers for specific embedded tuners in set-top boxes use the old (very old) structure.

> I could explain 24, but not 23...
> The device includes 2 PCIex modules:
>  . 1 module providing 8 DVB-S tuners
>  . 1 module providing 8 tuners for either DVB-T or DVB-C
> 3x8 = 24... but not 23

When a tuner provides DVB-T and DVB-C, then either you have two frontends, one for each modulation, or one single fontend and you need to specify the right "delivery system" when you tune. But, in all cases, the number of demux and dvr are the same as the number of frontends.

This can be an issue with TSDuck. In the `tslsdvb` output you provided, the frontends 0 to 7 are identified as DVB-C and 8 to 15 as DVB-S. There is no DVB-T. There is something to analyze here.

In the last code:
~~~
    UStringVector names;
    ExpandWildcard(names, u"/dev/dvb/adapter*");

    // Android
    ExpandWildcardAndAppend(names, u"/dev/dvb*.frontend*");

    // Linux
    ExpandWildcardAndAppend(names, u"/dev/dvb/adapter*/frontend*");
~~~
Do you still need the first one? It seems redundant with the third one.

And the comments are probably not adequate. It does not seem to be a matter of Linux vs. Android but rather a matter of legacy for the flat structure.

lelegard:
> This can be an issue with TSDuck. In the tslsdvb output you provided, the frontends 0 to 7 are identified as DVB-C and 8 to 15 as DVB-S. There is no DVB-T. There is something to analyze here.

To give you an example of a combined DVB-T / DVB-C tuner, here is the output of an AverMedia TD 310. You can see that the device is identified as "DVB-T, DVB-T2, DVB-C". There is only one tuner but it supports 3 different delivery systems. Use `-I dvb --delivery-system DVB-2` for instance to select one.

Note that, on Windows, the same device appears as two distinct tuners, a DVB-T/T2 one and a DVB-C one. I think that this is due to some limitation in the DirectShow framework. Since there is only one physical tuner anyway, the "two logical tuners" on Windows cannot be used at the same time.
~~~
$ tslsdvb -v

/dev/dvb/adapter0 (Silicon Labs Si2168, DVB-T, DVB-T2, DVB-C)

tslsdvb: error reading status on /dev/dvb/adapter0/frontend0: Resource temporarily unavailable
  Delivery system ............................ DVB-T 
  Frequencies:
    Min ................................. 48,000,000 Hz
    Max ................................ 870,000,000 Hz
    Step .................................... 62,500 Hz
    Tolerance .................................... 0 Hz
  Spectral inversion ........................... off 
  Modulation ................................... QAM 
  Bandwidth ................................... auto 
  FEC(high priority) .......................... auto 
  FEC(low priority) ........................... auto 
  Transmission mode ........................... auto 
  Guard interval .............................. auto 
  Hierarchy ................................... auto 

  Capabilities: inversion auto, FEC 1/2, FEC 2/3, FEC 3/4, FEC 5/6, FEC 7/8,
    FEC auto, QPSK, 16-QAM, 32-QAM, 64-QAM, 128-QAM, 256-QAM, QAM auto,
    transmission mode auto, guard interval auto, hierarchy auto, multistream,
    2nd generation, mute TS
~~~

SS-FF:
> Do you still need the first one? It seems redundant with the third one.

I apologize for the inaccuracy. I'm stuck at an earlier version (before introduction of the "guts"), so I'm just pasting the relevant changes into the GitHub web editor..
 
> And the comments are probably not adequate. It does not seem to be a matter of Linux vs. Android but rather a matter of legacy for the flat structure.

Fixed ;-)

SS-FF:
> > This can be an issue with TSDuck. In the tslsdvb output you provided, the frontends 0 to 7 are identified as DVB-C and 8 to 15 as DVB-S. There is no DVB-T. There is something to analyze here.

I don't think it's worth the effort. It's a closed system, not meant for use by applications, and as such it's sufficient that it makes sense to themselves. Maybe there's a jumper on a circuit board to switch to DVB-T, maybe a certain ioctl command or whatever it may be - it doesn't really matter.

> To give you an example of a combined DVB-T / DVB-C tuner, here is the output of an AverMedia TD 310. You can see that the device is identified as "DVB-T, DVB-T2, DVB-C". There is only one tuner but it supports 3 different delivery systems. Use `-I dvb --delivery-system DVB-2` for instance to select one.
> 
> Note that, on Windows, the same device appears as two distinct tuners, a DVB-T/T2 one and a DVB-C one. I think that this is due to some limitation in the DirectShow framework. Since there is only one physical tuner anyway, the "two logical tuners" on Windows cannot be used at the same time.

It's not a technical limitation. but a historical one: Some applications (including WMC) didn't provide a way to select a delivery system, and a device manufacturer wants to be compatible with all applications. That's my interpretation why it has evolved like that..

PR #623, opened 2020-08-07
Fix mn typo

rubenochiavone:
Broadcast time reported by tsanalyze command contains a typo in the minutes - mn instead of min.


lelegard:
Alright. Initially, it was not a typo, "mn" was intentional since it is a usual abbreviation for minute. However, according to Wikipedia, the SI symbol for minute is "min", not "mn". I respect that. I will merge it.

rubenochiavone:
@lelegard as you mentioned that "mn" was not a typo, sorry for that, maybe I can change the commit message to be more appropriate in that sense, what do you think?

lelegard:
> @lelegard as you mentioned that "mn" was not a typo, sorry for that, maybe I can change the commit message to be more appropriate in that sense, what do you think?

@rubenochiavone, don't worry about the message, only the code is important. 
Merged.

PR #676, opened 2020-12-03
Improve interrupt handling on Windows and allow termination via text …

SS-FF:
This PR includes three closely related changes for shutdown handling on Windows:

### 1. Fix CTRL-C Handler not working

To reproduce:

- Run tsp.exe from the command line
  (without any arguments)
- Press CTRL-C
=> tsp.exe does not exit and the only way to get out of this is to kill the tsp process

### 2. Process Close, Logoff and Shutdown system events

There are DVB devices where the driver causes a bluescreen when an active receive operation is not properly shut down.
That means: killing the tsp process == BSOD

When a user session is closed (by logoff, system poweroff or reboot), Windows sends the corresponding events (CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT, CTRL_SHUTDOWN_EVENT) to all processes. Those that haven't shut down gracefully after 5 seconds are getting killed - which this PR prevents from happening.

### 3. Allow ending the process via keyboard input

When you start tsp.exe from another process rather than manually from the command line, there is no way to stop the tsp process gracefully. Unfortunately it's practically impossible on Windows, to send a CTRL-C signal to a single child-process. I had extensively researched this a year ago and it's simply a hard (and sad) fact. Remaining workarounds are coming with an unacceptable set of drawbacks (hence I'm saying 'practically impossible').
I can provide more details on the subject in case anybody would be interested.

To provide an easy way for allowing a graceful shutdown in those cases, this PR adds reading of key stroke input for Windows and when a 'q' or 'Q' is entered, operations are shut down in the same way as with a CTRL-C signal.

This should not interfere with command lines where multiple processes are piped together, because the `kbhit()` function will return zero (0) in those cases. The input key reading is mostly adopted from ffmpeg which is capable to do both of these things as well.

Having a console interface would open up a range of other nice use cases as it provides a simple way of interaction, no matter whether it's used interactively from a console or from a parent process. For example, it could be used in a way that you can press certain keys to request printing of certain statistics while running, or retrieving the current signal strength and quality values from a running dvb input. And of course - live manipulation of certain parameters.

But for this PR, it's just about having a simple way to initiate a graceful shutdown.

### DIFF
--- a/src/libtsduck/base/tsUserInterrupt.cpp
+++ b/src/libtsduck/base/tsUserInterrupt.cpp
@@ -30,6 +30,11 @@
 #include "tsUserInterrupt.h"
 #include "tsSingletonManager.h"
 #include "tsSysUtils.h"
+
+#if defined(TS_WINDOWS)
+    #include <conio.h>
+#endif
+
 TSDUCK_SOURCE;
 
 ts::UserInterrupt* volatile ts::UserInterrupt::_active_instance = nullptr;
@@ -104,6 +109,100 @@ void ts::UserInterrupt::main()
 }
 #endif
 
+//----------------------------------------------------------------------------
+// Key input handler thread on Windows platforms
+// currently used for graceful app termination.
+//----------------------------------------------------------------------------
+
+#if defined(TS_WINDOWS)
+
+static int read_key(void)
+{
+    unsigned char ch;
+#if 0
+    // Linux equivalent. Requires termios.h
+    int n = 1;
+    struct timeval tv;
+    fd_set rfds;
+
+    FD_ZERO(&rfds);
+    FD_SET(0, &rfds);
+    tv.tv_sec = 0;
+    tv.tv_usec = 0;
+    n = select(1, &rfds, NULL, NULL, &tv);
+
+    if (n > 0) {
+        n = read(0, &ch, 1);
+        if (n == 1) {
+            return ch;
+        }
+
+        return n;
+    }
+
+#elif defined(TS_WINDOWS)
+    static bool is_pipe;
+    static HANDLE input_handle;
+    DWORD dw, nchars;
+
+    if(!input_handle){
+        input_handle = GetStdHandle(STD_INPUT_HANDLE);
+        is_pipe = GetConsoleMode(input_handle, &dw) == 0;
+    }
+
+    if (is_pipe) {
+
+        /* When running under a GUI, you will end here. */
+        if (!PeekNamedPipe(input_handle, nullptr, 0, nullptr, &nchars, nullptr)) {
+            // input pipe may have been closed by the parent process
+            return -1;
+        }
+
+        //Read it
+        if(nchars != 0) {
+            read(0, &ch, 1);
+            return ch;
+        }
+
+        return -1;
+    }
+
+    if(kbhit()) {
+        return(getch());
+    }
+
+#endif
+    return -1;
+}
+
+void ts::UserInterrupt::main()
+{
+    while (!_terminate) {
+
+        // Read key input
+        const int key = read_key();
+
+        if (key == 'q' || key == 'Q') {
+
+            std::cerr << "Received Quit key command" << std::endl;
+            std::cerr.flush();
+
+            // Set interrupted state
+            _interrupted = true;
+            _terminate = true;
+
+            // Notify the application handler
+            if (_handler != nullptr) {
+                _handler->handleInterrupt();
+            }
+
+            break;
+        }
+
+        SleepThread(50);
+    }
+}
+#endif
 
 //----------------------------------------------------------------------------
 // Handler on Windows platforms. Invoked in the context of a system thread.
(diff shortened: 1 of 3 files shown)


lelegard:
Thanks, it is interesting.

> To provide an easy way for allowing a graceful shutdown in those cases, this PR adds reading of key stroke input for Windows and when a 'q' or 'Q' is entered, operations are shut down in the same way as with a CTRL-C signal.
>
> This should not interfere with command lines where multiple processes are piped together, because the kbhit() function will return zero (0) in those cases. The input key reading is mostly adopted from ffmpeg which is capable to do both of these things as well.

I do not understand the code in that context.

First, it seems that the background task is looping on a 50 ms polling (`read_key()` seems non-blocking). Or dit I miss something? Polling is honestly never an acceptable solution, unless you do not have any other acceptable solution for an essential feature.

In `read_key()`, when the standard input is a pipe, you read a character:
~~~
    if (is_pipe) {

        /* When running under a GUI, you will end here. */
        if (!PeekNamedPipe(input_handle, nullptr, 0, nullptr, &nchars, nullptr)) {
            // input pipe may have been closed by the parent process
            return -1;
        }

        //Read it
        if(nchars != 0) {
            read(0, &ch, 1);
            return ch;
        }

        return -1;
    }
~~~
So, if the input is a piped transport stream, you pick up one byte from the TS, right?

Moreover, the code seems to consider that the file is either a pipe or a console. But it can be a file (eg. `tsp <in.ts ...`).

> Having a console interface would open up a range of other nice use cases as it provides a simple way of interaction, no matter whether it's used interactively from a console or from a parent process. For example, it could be used in a way that you can press certain keys to request printing of certain statistics while running, or retrieving the current signal strength and quality values from a running dvb input. And of course - live manipulation of certain parameters.

For this type of features, `tspcontrol` is supposed to do the work. It can be enhanced if necessary.

Additionally, in `tsp.cpp`, why do you disable the "one-shot" parameter? I used one-shot on purpose: hit Ctrl-C, give a chance to the application to handle it gracefully, but if it does not then the next Ctrl-C should really kill the process. Maybe the system handling of the second Ctrl-C was not working on Windows despite the explicit deactivation but we have to understand why.

SS-FF:
> Additionally, in tsp.cpp, why do you disable the "one-shot" parameter? I used one-shot on purpose: hit Ctrl-C, give a chance to the application to handle it gracefully, but if it does not then the next Ctrl-C should really kill the process. Maybe the system handling of the second Ctrl-C was not working on Windows despite the explicit deactivation but we have to understand why.

Correct. The first CTRL-C deactivates the interrupt handler and nothing else happens.

lelegard:
> >  Additionally, in tsp.cpp, why do you disable the "one-shot" parameter? I used one-shot on purpose: hit Ctrl-C, give a chance to the application to handle it gracefully, but if it does not then the next Ctrl-C should really kill the process. Maybe the system handling of the second Ctrl-C was not working on Windows despite the explicit deactivation but we have to understand why.
>
> Correct. The first CTRL-C deactivates the interrupt handler and nothing else happens.

It is anyway a Windows specific issue. On Linux and macOS, when the application does not terminate in response to the first Ctrl-C, the second one kills the application. We need to preserve this behaviour. Why doesn't it work on Windows, despite the deactivation code?

SS-FF:
> In `read_key()`, when the standard input is a pipe, you read a character:

> So, if the input is a piped transport stream, you pick up one byte from the TS, right?

This code does not handle cases where the input is a pipe or when commands are piped together.

> /* When running under a GUI, you will end here. */

This is handling cases when running tsp.exe as a child process from another application.

> Moreover, the code seems to consider that the file is either a pipe or a console. But it can be a file (eg. tsp <in.ts ...).

Neither with pipe input nor with file input should the code read anything. (and doesn't)

SS-FF:
All that logic is copied from ffmpeg. I wanted something that is known to work reliably instead of going through a lot of trial and error.

lelegard:
> This code does not handle cases where the input is a pipe or when commands are piped together.

I cannot merge code which would break the input transport stream when it comes from a pipe. This is a very common use case, either using the `fork` plugin or on Unix systems in general.

Interestingly, the automatic CI workflow of your PR fails on Windows only, in tests 66 and 70, which are precisely tests where the `tsp` input is a pipe from the `fork` plugin (see [here](https://github.com/tsduck/tsduck/pull/676/checks?check_run_id=1490478690)). So, the PR indeed breaks the code.

SS-FF:
I couldn't make much sense out of these logs. Can I run this locally?

SS-FF:
Maybe I'm missing some bit that ffmpeg uses to switch between those cases (console vs. piped input).

lelegard:
> I couldn't make much sense out of these logs. Can I run this locally?

Yes, you just need to run the test suite from bash, typically a "git bash" console on Windows. 

Have a look at https://github.com/tsduck/tsduck-test. Just clone the tsduck and tsduck-test projects side by side in the same parent directory. Compile the tsduck project, either using `build.ps1` or VS IDE. Then, in the tsduck-test project, run:
~~~
./run-all-tests.sh --dev
~~~
Or `--dev32`, `--debug`, `--debug32`, depending on which build you want to test. Or `--help`.

Because of the nature of processes and file systems, running the complete test suite takes 5 times more time on Windows than Linux (five minutes vs. one).

lelegard:
And, more generally, [testing TSDuck is described here](https://tsduck.io/doxy/testing.html).

You have the `utest` binary in the tsduck project to run the unitary test suite (individual C++ code features) and the tsduck-test project for application-level test suite.

SS-FF:
Thanks, I'll try. 

PS: I'm currently in the process of re-applying all my changes to the latest tsDuck version (after two rebase-attempts had ended up in a total mess). So, the code I'm submitting is what I've been using for 2 years now and I don't remember all details.

lelegard:
Hi @SS-FF 

I have come back on the Ctrl-C issue with a wider approach, not limited to the Windows implementation is `UserInterrupt` and I found an interference with a wider locking issue during the startup phase of `tsp` (before receiving the first TS packet from the input plugin). I fixed this issue. Now, a Ctrl-C on Windows during the startup phase correctly terminates the process. After the startup phase, there was no issue.

So, concerning your 3 topics:

1. Fix CTRL-C Handler not working
    - Fixed (please test).
2. Process Close, Logoff and Shutdown system events
    - I have included CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT, CTRL_SHUTDOWN_EVENT in the set of events (easy).
3. Allow ending the process via keyboard input
    - This one seems inappropriate since using an input pipe is a very common use case with `tsp` and `tspcontrol` already provides an equivalent feature.

SS-FF:
> Fix CTRL-C Handler not working
> 
>    Fixed (please test).

I'm afraid but I need all three things working.

> Allow ending the process via keyboard input
> 
>     * This one seems inappropriate since using an input pipe is a very common use case with `tsp` and `tspcontrol` already provides an equivalent feature.

Please note, that on Windows, it is practically impossible to send CTRL-C to a child console application process from a Windows application. Establishing an additional network connection to TSP just for being able to end the process is not an acceptable solution IMO.

I would envision for TSP to behave similar to ffmpeg. ffmpeg can do both: input piping and key input control. It would be great when the tspcontrol communication would be available via STDIN and STDERR as well. (STDERR because this way, output piping will still be possible - ffmpeg does it like that).

It's fine when you have other plans, though.

Thanks for addressing 1 and 2!

PR #681, opened 2020-12-07
Adjust DEFAULT_SINK_QUEUE_SIZE

SS-FF:
Now that we know that the allocator provides a pool of only 24 media samples, we should adapt this to avoid the upstream filters from running out of available samples.

### DIFF
--- a/src/libtsduck/dtv/tsTuner.h
+++ b/src/libtsduck/dtv/tsTuner.h
@@ -317,7 +317,7 @@ namespace ts {
         //! Default max number of queued media samples (Windows-specific).
         //! @see setSinkQueueSize().
         //!
-        static constexpr size_t DEFAULT_SINK_QUEUE_SIZE = 1000;  // media samples
+        static constexpr size_t DEFAULT_SINK_QUEUE_SIZE = 12;  // media samples
 
         //!
         //! Set the max number of queued media samples (Windows-specific).


lelegard:
Hi @SS-FF 

I do not think that this is appropriate to set a limit of 12 media samples when the framework can allocate up to 24. There is a risk of dropping packets for nothing.

This limit is just a foolproof check. It was done only to make sure that the driver does not allocate an infinite number of media samples while the application reads nothing, leading to a memory exhaustion. Specifically, we do not allocate anything physical with that limit (no physical queue with 1000 unused entries).

So, it just important to use some finite value, something higher than the possible typical maximum number of media samples in the push/pull adaptation queue (ie. higher than 24) and lower than a crazy value that would exhaust the memory.

We do not really know if 24 is always used by all tuners in all versions of Windows. So, we should use something safe, I would say 100, 1000 or 2000 are all valid values. 12 is too small and 1000000 would be too high. I see no reason to change the current value.

SS-FF:
As soon as you have 24 samples in the queue, the upstream filter is no longer able to get a free sample to write to.
And THEN you will loose packets. So anything >= 24 is wrong. Practically, anything close to 24 is problematic.

Instead of hard-coding, you can also use the following code to determine the number of samples that the allocator provides:

```cpp
STDMETHODIMP ts::UdpSinkPin::NotifyAllocator(IMemAllocator* pAllocator, BOOL bReadOnly)
{
    TRACE(1, u"UdpSinkPin::NotifyAllocator");

    ALLOCATOR_PROPERTIES props;
    pAllocator->GetProperties(&props);

    std::cerr << "UdpSinkPin::NotifyAllocator - buffers: " << props.cBuffers << " buffer size: " << props.cbBuffer  << std::endl;
    std::cerr << "UdpSinkPin::NotifyAllocator - cbAlign: " << props.cbAlign << " cbPrefix: " << props.cbPrefix  << std::endl;

    return S_OK; // we don't care
}
```

And then use  ` props.cBuffers / 2`  to set the queue size.

SS-FF:
As an alternative, copy the data immediately and don't `.AddRef` the media samples. But as we already discussed, this would require an additional copying or an architectural change.

lelegard:
> As an alternative, copy the data immediately and don't .AddRef the media samples. But as we already discussed, this would require an additional copying or an architectural change.

As already discussed, that would be counter-productive. Those 24 media samples with 240 packets each (as you already reported) represent from 0.1 (DVB-S2) to 0.3 seconds (DVB-T). Any streaming application which cannot read 10 times per second won't work anyway, not only because it does not read fast enough, but because it would be badly architectured on a real-time perspective (come one, 10 times per second is nothing). And that is not the case with `tsp`. So, adding a level of intermediate buffering can only makes things worse and that won't be done.

SS-FF:
But how do you want to address it then? When you keep 24 samples allocated, the filter graph cannot continue and that can lead to packets getting lost upstream. Didn't you and @oviano talk about having experienced packet corruption with DirectShow?

oviano:
So if I understand this correctly, by having a large value here, such as the default, and then not reading quickly enough in ts::Tuner::receive(), this could cause DShow to drop packets further up the graph because there would be nowhere to place them (limited to 24 samples)?

Whereas with a lower value, such as 12, what would happen then?

SS-FF:
> So if I understand this correctly, by having a large value here, such as the default, and then not reading quickly enough in ts::Tuner::receive(), this could cause DShow to drop packets further up the graph because there would be nowhere to place them (limited to 24 samples)?

Correct. But it depends on what's happening upstream. As soon as tsDuck has those 24 samples in the queue, the upstream filter won't be able to get a sample from the allocator.
That does not immediately cause a loss of packets, because the the `IMemAllocator::GetBuffer()` method blocks (https://docs.microsoft.com/en-us/windows/win32/api/strmif/nf-strmif-imemallocator-getbuffer) until a sample will be available again. Once we look at one more level upstream, we'll get to the connection between the hardware filter where the data is transmitted from kernel- to user-mode via the kernel streaming proxy (ksproxy). 
This connection may have a different allocator, but we can't and shouldn't make any assumptions about it. Additionally, there might some buffering at the driver side, but again, we shouldn't make any assumptions about it.

There surely is some margin for buffering happening upstream - otherwise we would have seen much bigger problems already.
But we can't and shouldn't rely on this and instead process the incoming packets in a way that it doesn't cause stalling of the data flow in the filter graph.

Of course it's a dilemma about where the stalling might occur. But when we know that we can have a maximum of 24 media samples, why choose a queue size of 1000. Why 1000, when all we can ever get is 24?

When all we can ever get is 24, and we don't want to do an additional intermediate copying, then we need to make sure instead that the data is further processed by TSP at a sufficient frequency. 

> 
> Whereas with a lower value, such as 12, what would happen then?
> 
> I struggle to see why my loop would not be able to keep up, it’s one thread that basically sits on receive() all the time, with a timeout set to 50ms. Outside of the receive() call it’s doing almost nothing. Checking a couple of flags to know whether to quit, or restart is about all it does. Any data it receives is simply passed to another thread to deal with demuxing etc so there is no chance of it blocking.

The problem is that I don't see those packet corruptions, so I can't say for sure whether it can be resolved this way.
It would be great when you could just give it a try.. 😄

lelegard:
OK, I think it is time to make things clear:

1. Packet corruption, as we have seen them sometimes, have nothing to do with buffer shortage. If there is no buffer, there is no packet, not corrupted packets.

2. 24 media samples with 240 packets in each (as reported by @SS-FF) contain 8 Mb, corresponding to 0.1 second of content in the worst case. In those conditions, a buffer shortage is an application problem, nothing else. A normal application will release the media samples fast enough for reuse.

3. Resizing the SinkFilter queue to 12 will make things worse because the pin Receive will block and packets will be lost while there are still 12 additional free media samples which could be used.

4. Copying in intermediate buffers and early release of media sample will only solve packet losses for _bad_ applications which are too slow, while penalising _good_ applications by adding an additional buffer management and packet copy.

So, unless there is an actual _proof_ that there is an issue here, let's stop speculations.

SS-FF:
From my experience, a _good_ DirectShow application doesn't holds references to **IMediaSample** objects that are received by an input pin. As simple as that.
(exceptions exist for certain scenarios)

Regarding the specific effects and behavior regarding the buffer size parameters in case of the current tsDuck implementation, I agree that we're currently in an area of speculation. That's why I have suggested @oviano to **try**.

Meanwhile I got feedback from 4 testers, all reporting that things (channel scanning, recording+playback) are still working fine (SinkFilter queue set to 12).  Of course that's no evidence for whether it's an improvement.

The most important change has been setting the initial_packet_count to 10 which was kind of game changer for me and allowed me to move away from the UdpSink filter I had done, because the timing advantage was < 50 ms after specifying initial_packet_count. Also, I'm not seeing the initial packet loss (after startup) anymore.

> Packet corruption, as we have seen them sometimes, have nothing to do with buffer shortage. If there is no buffer, there is no packet, not corrupted packets.

Only as long as the upstream implementation doesn't lose packet boundary alignment in case of buffer shortage.

lelegard:
> From my experience, a good DirectShow application doesn't hold references to IMediaSample objects that are received by an input pin. As simple as that.

I totally disagree with that statement. This is just wrong. You even provided in a previous post (can't find it again), a link to a Microsoft article which explicitly mentioned several strategies to handle a media sample. The first one was to process it immediately in the pin Receive(). The other one was to send the media sample to a "worker thread" for later processing, this worker thread being responsible for releasing the media sample later. You may have a good experience with the first strategy but the second one is perfectly valid as well and anticipated in the design of DirectShow (as stated in the MS article).

> Meanwhile I got feedback from 4 testers, all reporting that things (channel scanning, recording+playback) are still working fine (SinkFilter queue set to 12). Of course that's no evidence for whether it's an improvement.

So, this clearly means that the application (here, `tsp` I assume) is fast enough to collect samples with only a depth of 12. This means that an additional intermediate buffering to release media sample earlier is useless.

lelegard:
Here is the link: https://docs.microsoft.com/en-us/windows/win32/api/strmif/nf-strmif-imeminputpin-receive

which says:

> If the pin uses a worker thread to process the sample, it holds a reference count on the sample. In any case, the output pin cannot directly re-use this sample. It must call the IMemAllocator::GetBuffer method to obtain a new sample.

SS-FF:
> > From my experience, a good DirectShow application doesn't hold references to IMediaSample objects that are received by an input pin. As simple as that.
> 
> I totally disagree with that statement. This is just wrong. You even provided in a previous post (can't find it again), a link to a Microsoft article which explicitly mentioned several strategies to handle a media sample. The first one was to process it immediately in the pin Receive(). The other one was to send the media sample to a "worker thread" for later processing, this worker thread being responsible for releasing the media sample later. You may have a good experience with the first strategy but the second one is perfectly valid as well and anticipated in the design of DirectShow (as stated in the MS article).

I said that there are exceptions as I didn't want to get into too many details. There are just many different possible situations. First of all: when you don't have any live source filter, you never need to be concerned about any kind of loss. Then, a DirectShow graph can be blocked forever at any point in the graph.
A typical case where you would want to keep a reference to a media sample are In-Place-Transform filters. Simple example would be a filter for audio volume adjustment. Samples could be add-refed and processed by another thread, then pushed downstream through an output pin. But even in a scenario like that, you wouldn't let the samples accumulate - you would implement this in a way to keep the data flowing through the graph.

SS-FF:
>  keep the data flowing through the graph.

We also shouldn't forget that we (the sink filter) are a child of the Tee Splitter filter and when we block the data flow, the other child(ren) - in this case, the "MPEG2 Demultiplexer" won't receive more data as well.

And the Network Provider filter communicates with the TIF filter (child of the MP2 Demux filter) to complete the tuning process.

SS-FF:
The above doesn't seem to be practically relevant in our case - it's just for illustration.

oviano:
Ideally what I would like to do, at some point, and whether this is through butchering @lelegard's beautifully engineered TSDuck code or starting afresh, is end up with a situation where at the point that DirectShow graph gives me some samples I can just copy them directly into my buffer for my demux thread and skip the whole intermediate queue, avoiding adding any references completely.

I'm making this up as I go along without studying it more carefully but I wonder if we might be able to have some sort of virtual function in ts::Tuner that gets called when data is pushed in this manner, and then the application can do what it sees fit with it.

A sort of alternative push mechanism, hell maybe even Windows-only, that sidesteps the queue and is for use by evil hackers like me :)

SS-FF:
I have two ideas, but I need to tell that both are not really thought through in all detail.

The first one is maybe even a bit insane...

#### Idea 1

The TSP calls the tuner's receive method:

```
size_t receive(TSPacket* buffer, TSPacketMetadata* pkt_data, size_t max_packets)
```

Assuming that for example, max_packets is 1000 and the method returns 500.
Couldn't the tuner store the buffer pointer plus the information that there is a space of 500 packets left?

Then the sink filter could directly copy incoming packet data into that remaining space... 

#### Idea 2

This is more a kind of question which I'm wondering about. Assuming we would have an intermediate buffer in the sink filter - it could be large enough to hold a number of contiguous samples.

And then, for example - as each sample arrives, it will be copied to that larger buffer. When the tuner::Receive method is called, it would take just a single memory copy of the contiguous intermediate buffer.

The question where I'm not sure about - assuming 12 IMemorySamples would arrive

- we will always need to perform 12 copy operations
- in case of having an intermediate buffer, there would be a single additional copy operation (just a larger segment of memory top copy)

How much would that additional single copy operation weigh against the other 12?

oviano:
So in the end I decided to put on my dirty evil hacking gloves and experiment with directly pushing the data from DirectShow to my application code.

It's here if anyone is interested:

https://github.com/oviano/tsduck/tree/directshow-push-api

Essentially I derive a custom tuner class from ts::Tuner and it's push() gets called with the data from the samples directly. The assumption is the data is only valid for the duration of the call, so the push() function implementation would be expected to copy it or use it but not rely on the data existing any longer. In my case, this allows me to more elegantly push it straight into my libavformat demux buffer.

Obviously this is a Windows-only change, but it will at least allow me to determine whether taking a more direct approach has any bearing on reliability/packet loss/corruption etc.

SS-FF:
> Obviously this is a Windows-only change, but it will at least allow me to determine whether taking a more direct approach has any bearing on reliability/packet loss/corruption etc.

I'm curious about your results. Please let us know about your findings..

oviano:
Anecdotally it seems more stable, but I’m testing on a somewhat flakey DVB-T signal using an indoor antenna so it can be difficult to differentiate between signal fluctuations and other corruption. Also, since I am now bypassing the queue I’m not running that chunk of code that searches for bounding 0x47 bytes and re-syncs on a discrepancy, logging as it does so so I’m rather running blind. I’m simply throwing what I receive directly from directshow at the libavformat demuxer to deal with.

@lelegard is there any form at all you would like this feature in TSDuck? Clearly it benefits none of the existing code and plugins and would be a somewhat hidden API for users of tsducklib and maybe nobody would ever use it apart from me but if there was some form you wanted it in I’d be happy to change it.

At the moment to use it you derive from ts::Tuner and override two virtual functions 1) to tell the sink filter that you will handle the data being pushed at you and 2) to receive the pushed data. If you don’t override these functions it’s normal behaviour and pushes data to the queue as normal. I could at least cache the result of the first function if the virtual overhead bothered you. But maybe there is a more elegant or preferable way you would want it to work, perhaps by passing in a new class object to receive the data rather than derive from ts::Tuner (a bit like how you provide a report handler for logging, I suppose).

Or more likely, this whole idea sucks in your view and you don’t want anything to do with it, which is fine too and I’ll just keep my fork and periodically merge your latest :)

lelegard:
> @lelegard is there any form at all you would like this feature in TSDuck? Clearly it benefits none of the existing code and plugins and would be a somewhat hidden API for users of tsducklib and maybe nobody would ever use it apart from me but if there was some form you wanted it in I’d be happy to change it.

This would introduce a Windows-only high-level feature and I do not like the idea. Currently, the TSDuck library has close to zero system-specific API, apart from tuning a pair of very specific low-level parameters which make sense on one system only.

For me, the atrocity of DirectShow has become a fact. My recommendation remains the same: if you want a reliable and future-friendly platform for tuners, use Linux. Tuner support on Windows will never be better than "best effort". If the result is good enough for you, fine. Otherwise, use Linux.

PR #682, opened 2020-12-07
Windows Tuners: Use the same polling for signal lock as for Linux

SS-FF:
Maybe use the same polling option as for Linux

### DIFF
--- a/src/libtsduck/dtv/windows/tsTunerGuts.cpp
+++ b/src/libtsduck/dtv/windows/tsTunerGuts.cpp
@@ -472,22 +472,33 @@ bool ts::Tuner::start(Report& report)
         return false;
     }
 
-    // If the tuner was previously started/stopped on a frequency with signal on it,
-    // it has been observed that remaining packets from the previous run were still
-    // there. Wait a little bit and reflush after Run() to avoid that.
-    // Yes, this is a horrible hack, but if you have a better fix...
-    SleepThread(50); // milliseconds
-    sink->Flush();
+    // Wait for input signal locking if a non-zero timeout is specified.
+    MilliSecond pollInterval = 100; // maybe use DEFAULT_SIGNAL_POLL and a common option for Linux and Windows
 
-    // If a signal timeout was specified, read a packet with timeout
-    if (_signal_timeout > 0) {
-        TSPacket pack;
-        if (sink->Read(&pack, sizeof(pack), _signal_timeout) == 0) {
-            if (!_signal_timeout_silent) {
-                report.error(u"no input DVB signal after %'d milliseconds", {_signal_timeout});
-            }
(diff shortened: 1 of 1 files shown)


lelegard:
Hi @SS-FF,

What you propose is in fact the original implementation of TSDuck tuner on Windows. Since the first TSDuck version was developed on Linux, the Windows implementation was more or less inspired from the original Linux implementation. But, after some time, I realized that some new tuners (or more precisely their proprietary Windows drivers) were unable to correctly report the state of the tuner or some parts of it. This includes current tuning parameters, state of signal (locked, quality, strength). So, with these tuners, the tuning operation "failed" because no lock state was ever reported.

As a last resort, I implemented the signal check as the read of one packet. It was clear that some tuners could actually read packets while reporting "no signal lock", "strength 0%" or "quality 0%".

On the other hand, the same tuner could correctly report its current tuning parameters and signal state on Linux, with the open source drivers.

The example which particularly frustrated me was the following. On DVB-T (unlike DVB-S), you don't really need to know all tuning parameters (FEC, guard interval, etc). As long as you know the frequency, just tune and the tuner will adapt. Even better, at least on Linux, when you read back the tuning parameters, you get the actual parameters. This is very handy when you need to draw a map a DVB-T network characteristics (a task that was useful to me at the time). But, after porting the tuner part of TSDuck to Windows to move to various areas using my WIndows laptop, I soon realized that the same tuner was unable to report the actual parameters on Windows. It just returned the dummy / default parameters I gave on input. This was of course a driver issue because the same information was available on Linux. And that was the case for _all_ tuners I tested on Windows. Note that this is the reason why I developed the ability for `tsterinfo` to "guess" the tuning parameters based on the bitrate. The first tuners I used on Windows were however able to report "signal lock". But, after testing some others, I realized that was not the case for all of them and that "signal locked" was, at best, not a trustable information on Windows.

Since this was a long time ago, we are tempted to say that the situation should be better now. This is unfortunately the opposite. The bad driver implementation has event contaminated the Linux area. Now, even some Linux drivers can't report the signal quality properly to such extent that I had to simply disable the feature on Linux (report "not implemented" status).

The lessons I learned is that it is dangerous to rely on the behaviour of one specific tuner to change the code, on Windows from the beginning and, quite unfortunately, more and more on Linux. The current code is the result of many changes based on my or other user's experience (which includes yours for some other parts).

So, I am sorry, but I cannot replace the "read one initial packet" by "read signal locked", it won't work on some tuners.

SS-FF:
The reporting of actual tuning parameters and signal-locked state are two different things.
The signal-locked state seems to be an intrinsic feature because it is also used for the event mechanism with the `ITuner::TriggerSignalEvents` method.

Anyway, I'll report back once I got some feedback from our testing group.

PR #683, opened 2020-12-07
Avoid choosing tuner from enumeration when a device path is specified

SS-FF:
I've seen cases where the tuner device enumeration can take quite some time. This PR avoids the enumeration and creates the moniker directly from a device path.


lelegard:
This PR was merged. Github does not show it as such probably because of the previous rebase. I close it but, do not worry, it is merged.

PR #691, opened 2020-12-11
Fix bug in tsTunerGraph

SS-FF:
The code path 

`added receiver filter \"%s\" - but need additional filter` 

has never been functional, because cleanupDownstream was always cleaning up everything after the tuner filter - which also included a receiver filter that needs an additional filter.


SS-FF:
Sorry, this was about one of my own changes, covering cases where more than two tuner filters need to be added.

PR #704, opened 2020-12-23
tsscan: Display modulation parameters for ISDB-T broadcasts.

nicst35:
#### Related issue (if any):
The issue related with incorrect transmission parameters with some ISDB-T tuners (last posts on #24)

#### Affected components:
`tsscan` and libtsduck/dtv/tsModulationArgs.cpp

This series of commits provide a fix for the invalid modulation parameters with some ISDB-T tuners.
Basically this solution:

- Uses TSScanner to retrieve the modulation parameters that should be displayed.
- Inside TSScanner, do the tuning parameters query as late as possible to give the device some time to update its internal status.
- Corrects the enumeration used for displaying per-layer constellation types.
- Checks the validity of some parameters before showing them.
- Fixes some typos with the displayed data.

### DIFF
--- a/src/libtsduck/dtv/tsModulationArgs.cpp
+++ b/src/libtsduck/dtv/tsModulationArgs.cpp
@@ -1030,6 +1030,10 @@ void ts::ModulationArgs::display(std::ostream& strm, const ts::UString& margin, 
         strm << margin << "Modulation: " << ModulationEnum.name(modulation.value()) << std::endl;
     }
 
+    if (!delivery_system.set()) {
+        return;
+    }
+
     switch (TunerTypeOf(delivery_system.value())) {
         case TT_DVB_C: {
             if (symbol_rate.set() && symbol_rate != 0) {
--- a/src/libtsduck/dtv/tsModulationArgs.cpp
+++ b/src/libtsduck/dtv/tsModulationArgs.cpp
@@ -1151,10 +1151,10 @@ void ts::ModulationArgs::display(std::ostream& strm, const ts::UString& margin, 
             if (layer_a_modulation.set() && layer_a_modulation != QAM_AUTO) {
                 strm << margin << "Layers A modulation: " << InnerFECEnum.name(layer_a_modulation.value()) << std::endl;
             }
-            if (layer_a_segment_count.set()) {
+            if (layer_a_segment_count.set() && layer_a_segment_count.value() <= 13) {
                 strm << margin << "Layers A segment count: " << layer_a_segment_count.value() << std::endl;
             }
-            if (layer_a_time_interleaving.set()) {
+            if (layer_a_time_interleaving.set() && layer_a_time_interleaving != 255) {
                 strm << margin << "Layers A time interleaving: " << layer_a_time_interleaving.value() << std::endl;
             }
             if (layer_b_fec.set() && layer_b_fec != FEC_AUTO) {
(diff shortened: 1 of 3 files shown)


lelegard:
Hi @nicst35,

Thanks for your contribution on ISDB-T (including fixing a couple of silly mistakes of mine).

I have added a more formal check of the parameters. I was especially doubtful about the check `!= 255` on time interleaving values. The value 255 looked too specific to a given driver to be individually checked. It replaced this with a check of all possible documented values.

Could you please check that I didn't accidentally break something?

Could you also post an output of `tsscan` with the modified version, just like the one you posted in issue #24 ?

nicst35:
>The value 255 looked too specific to a given driver to be individually checked.

When I patched the driver, I noticed the following
```
struct sms_isdbt_layer_stats {
	/* Per-layer information */
	u32 code_rate; /* Code Rate from SMSHOSTLIB_CODE_RATE_ET, 255 means layer does not exist */
	u32 constellation; /* constellation from SMSHOSTLIB_CONSTELLATION_ET, 255 means layer does not exist */
	u32 ber; /* Post Viterbi ber [1E-5], 0xFFFFFFFF indicate N/A */
	u32 ber_error_count; /* Post Viterbi Error Bits Count */
	u32 ber_bit_count; /* Post Viterbi Total Bits Count */
	u32 pre_ber; /* Pre Viterbi ber [1E-5], 0xFFFFFFFF indicate N/A */
	u32 ts_per; /* Transport stream PER [%], 0xFFFFFFFF indicate N/A */
	u32 error_ts_packets; /* Number of erroneous transport-stream packets */
	u32 total_ts_packets; /* Total number of transport-stream packets */
	u32 ti_ldepth_i; /* Time interleaver depth I parameter, 255 means layer does not exist */
	u32 number_of_segments; /* Number of segments in layer A, 255 means layer does not exist */
	u32 tmcc_errors; /* TMCC errors */
};
```
And comparing to dvb_frontend.h (at kernel) we can see that the interleaving value is a unsigned 8-bit value.
```
struct dtv_frontend_properties {
        ...
	/* ISDB-T specifics */
	u8			isdbt_partial_reception;
	u8			isdbt_sb_mode;
	u8			isdbt_sb_subchannel;
	u32			isdbt_sb_segment_idx;
	u32			isdbt_sb_segment_count;
	u8			isdbt_layer_enabled;
	struct {
	    u8			segment_count;
	    enum fe_code_rate	fec;
	    enum fe_modulation	modulation;
	    u8			interleaving;
	} layer[3];
        ...
}
```
So yes, `255` is driver-specific. I simply casted `ti_ldepth_i` from `u32` to `u8` without any further modification or checking.

By the way, that value should map to a table defined at [kernel documentation](https://www.kernel.org/doc/html/latest/userspace-api/media/dvb/fe_property_parameters.html#dtv-isdbt-layer-a-c-time-interleaving). Setting it to 0 within the driver would be misreporting the value. Note that documentation says -1 can be a valid value but that seems to be a contradiction given the data types involved (i think).

nicst35:
> Could you also post an output of tsscan with the modified version, just like the one you posted in issue #24 ?

Sure, see [scan.txt](https://github.com/tsduck/tsduck/files/5752691/scan.txt)

Finally, due to all broadcasters in my area chose to use mode 3 according to the regulatory documents, I couldn't check what would the device return as `ti_ldepth_i` in other modes.

lelegard:
> So yes, 255 is driver-specific. I simply casted ti_ldepth_i from u32 to u8 without any further modification or checking.
>
> By the way, that value should map to a table defined at kernel documentation.

Yes, definitely. This is why I introduced the function `IsValidISDBTTimeInterleaving()` [here](https://github.com/tsduck/tsduck/blob/master/src/libtsduck/dtv/tsModulationArgs.h#L329) and [here](https://github.com/tsduck/tsduck/blob/master/src/libtsduck/dtv/tsModulationArgs.cpp#L229) to eliminate invalid values instead of checking `!= 255` just in case the kernel internals are changed in the future.

> Note that documentation says -1 can be a valid value but that seems to be a contradiction given the data types involved (i think).

Definitely. Since 0xFF is stored in a u8 and `ioctl(FE_GET_PROPERTY)` returns a list of u32, casting u8(-1) = 0xFF into u32 still holds 255, not -1. So, if the kernel comments say 0xFF means non-existent and the user API says -1 means auto, there is some inconsistency.

Do you plan to submit your patch of the Siano driver to the kernel? I think that would be very beneficial. I have noted that while long ago all Linux drivers could report current tuning parameters, recent drivers usually don't. This is a major regression and you patch definitely fights this global tendency to regression. No later than yesterday, when I tried `tsscan` on my local DVB-T network to check I did not break anything, I noticed that my old Hauppauge Nova-T correctly reported tuning parameters while my recent Avermedia TD310 didn't.

nicst35:
>Do you plan to submit your patch of the Siano driver to the kernel? I think that would be very beneficial

I sent it to the linux-media mailing list for review last week.

Issue #738, opened 2021-03-18
[Q] [SRT] Operation not supported: Bad parameters


Plombirius:
Hi @lelegard ,

We have discovered some issue when we have tried establish SRT connection between Rassberi Pi 4 ("Raspbian GNU/Linux 10 (buster)") and x86 Ubuntu ("Ubuntu 20.04.2 LTS").
When we have started receiver side we got error on sender side "* Error: srt: error during srt_accept(), msg: Operation not supported: Bad parameters" 
What did we wrong?

We started sender with this string: "`tsp -v --debug=1 -I dvb -a 0 --delivery-system DVB-T2 --frequency 498000000 --bandwidth 8-mhz -O srt 5555 --transtype live --messageapi`" and receiver -- "`tsp -v --debug=1 -I srt 192.168.0.242:5555 --transtype live --messageapi -O drop`"

root@raspberrypi:/home/pi# tsp --version
tsp: TSDuck - The MPEG Transport Stream Toolkit - version 3.26-2285

root@srt:/home/srt# tsp --version
tsp: TSDuck - The MPEG Transport Stream Toolkit - version 3.26-2289

libsrt version on both side - 1.4.3

Thank you in advance!

lelegard:
Hi @Plombirius,

Did you try with libsrt 1.4.2?

I see that libsrt 1.4.3 is not yet released (-rc.0 has been out 10 days ago only). I do not know if there is any change that could produce this behavior. Moreover, not being an SRT specialist, I cannot really validate the command.

Maybe @maxsharabayko could answer both questions, 1.4.2 vs. 1.4.3 and validity of the command.

lelegard:
Hi,

I did two quick tests, using available configurations I have.

- Linux Ubuntu 20.10 with libsrt 1.4.1 (coming with the distro), both ends on the same machine, same commands as yours. It works.
- Windows 10 with libsrt 1.4.2, both ends on the same machine. Same error as yours: "Error: srt: error during srt_accept(), msg: Operation not supported: Bad parameters".

I do not know if anything changed with SRT version 1.4.2. This looks more like an SRT question rather than a TSDuck question. We definitely need @maxsharabayko's assistance. If he does not answer here, try to report the issue in [Haivision/srt](https://github.com/Haivision/srt), they are aware of TSDuck.

lelegard:
Also tested on macOS 11 with libsrt 1.4.2 (installed with Homebrew), same error.

[...]

PR #742, opened 2021-03-21
Fix bad parameters issue on srt_accept()

spnngl:
A missing initialization on `int *addrlen` of srt_accept() cause a bad
paremeters error (SRT_EINVPARAM).

This error seems to have no consequences until libsrt >= 1.4.2

See: https://github.com/tsduck/tsduck/issues/738
See: https://github.com/Haivision/srt/blob/master/docs/API-functions.md#srt_accept


lelegard:
Well done, thanks, I merged it

PR #854, opened 2021-08-25
Update tsplugin_splicemonitor.cpp

wongfei2009:
refactor SpliceContext to get prepared for implementing preroll monitoring on time signal command

#### Affected components:

tsp {input} -P splicemonitor -O drop

Refactor SpliceContext so that it holds a map of splice event. It is necessary to support time signal command which can signal multiple events simultaneously.

### DIFF
--- a/src/tsplugins/tsplugin_splicemonitor.cpp
+++ b/src/tsplugins/tsplugin_splicemonitor.cpp
@@ -62,6 +62,18 @@ namespace ts {
         virtual Status processPacket(TSPacket&, TSPacketMetadata&) override;
 
     private:
+        // SCTE-35 splice event.
+        class SpliceEvent
+        {
+        public:
+            SpliceEvent();                // Constructor.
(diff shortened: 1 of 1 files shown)


lelegard:
Hi @wongfei2009 

Nice one, thanks.

Up to now, all use cases involved ads events being signaled a few seconds ahead. So, there were never simultaneously signaled events. But, in the generic case, you are right, it is theoretically possible to have simultaneously signaled events on the same PID if events are signaled a long time ahead.

I am curious to know which use case you have to need this.

wongfei2009:
Today, a lot of live feeds in US actually use SCTE35 time signal command rather splice insert to signal splice events. Very often, a single time signal command will carry multiple segmentation descriptor simultaneously (e.g. Program start & chapter start). 

I would like to enhance this splice monitoring plugin to be able to monitor the preroll time of time signal command. So, I try to enhance the SpliceContext firstly.

PR #869, opened 2021-09-11
Detect Centos Stream as minor version 99.

caveman99:
E.G. centos Stream 8 will be 899 All other releases contain proper minor revisions and thus igore the trailing .99

### DIFF
--- a/build/install-prerequisites.sh
+++ b/build/install-prerequisites.sh
@@ -181,7 +181,7 @@ elif [[ -f /etc/fedora-release ]]; then
 elif [[ -f /etc/redhat-release ]]; then
 
     # Red Hat or CentOS
-    EL=$(grep " release " /etc/redhat-release 2>/dev/null | sed -e 's/^.* release \([0-9]*\.[0-9]*\).*$/\1/')
+    EL=$(grep " release " /etc/redhat-release 2>/dev/null | sed -e 's/.*/&.99/' | sed -e 's/^.* release \([0-9]*\.[0-9]*\).*$/\1/')
     EL=$(( ${EL/.*/} * 100 + ${EL/*./} ))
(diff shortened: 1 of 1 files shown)


lelegard:
I may have misunderstood what you mean. How is CentOS stream exactly identified in `/etc/redhat-release`?

caveman99:
The Problem is, Centos Stream is only identified as 'CentOS Stream release 8' - so the second regex does not match anything and the following line bombs. The mod intends to appends .99 for the detection.

With regular point releasses the appended .99 is ignored.

Unfortunately there's no proper way to identify the point releases, as stream is a sliding target. Only the major release may change from 8 to 9 in the future...

caveman99:
Compilation just finished, so i can confirm that install-prerequisites.sh and the following build works with the patch at this point in time release of Centos 8 Stream, including Dektek SDK.

lelegard:
All right and it also works with `CentOS Linux release 8.4.2105`, the current content of a red-hot-release file.
I am going to merge this one.

lelegard:
I also need to check how are identified Alma Linux and Rocky Linux, the new clones of RHEL to replace CentOS. I doubt that CentOS Streams will gain any popularity. Nobody needs an "outdated unstable version". Either you need stability (RHEL, former CentOS, now Rocky and Alma Linux) or up-to-date software (Fedora). My company and all others we work with have plans to move away. Still unsure which one will be used but certainly not CentOS Streams, the operations guys require a stable OS.

caveman99:
All the other RHEL Clones (Oracle Linux and Amazon Linux 2 come to mind in addition to what you mentioned) are most likely using the Major.Minor system in /etc/redhat-release to be compatible. Stream is the only exception. As to your other points: we use the Stream distribution in a lab environment, since it basically turned into the the upstream for RHEL (and thus all other clones). Yes, you may need stability, and you may need the 'NEXT' release without leaving the EL environment and heading to Fedora. This patch does not affect detection adversely but adds another use option.

PR #875, opened 2021-09-28
Feature/javascript bindings

jkuemmele:
Javascript bindings to some TSDuck features.


lelegard:
Hi @jkuemmele 

Thanks for this original idea. It definitely opens new opportunities.

All automatic CI checks failed because "napi.h" is not available. I assume this is a new requirement for JavaScript bindings which must be installed before building. The scripts `build/install-prerequisites.sh` (Linux and macOS) and `build/install-prerequisites.ps1` (Windows) should be completed to make sure that all prerequisites are installed prior to building TSDuck.

The prerequisites shall be found for most Linux distros, macOS (using Homebre for opensource) and custom download & install PowerShell scripts on Windows. If one Linux distro or older version of a distro cannot provide the required support, this shall be detected in the makefile (see for instance how PCSC or SRT are managed in `Makefile.tsduck`). Then some symbol (typically `TS_NONAPI`) should allow conditional compilation when this API is not available.

I am not familiar with JavaScript environments (my background is Ada, C++, Java), so my questions may be naive.

There are two `yarn.lock` files, with "autogenerated" comments. Are they necessary? Shouldn't they be included in `.gitignore`? One of them contains references to specific tarballs with their sha512 value. Is this appropriate? Why is this used for?

More generally, for a given distro, are we able to generate a TSDuck binary package which is independent of the nodejs version which is installed on the target system? I ask this question because similar issues exist with Python but not with Java. In Java, the Java Native Interface (JNI) has been carefully designed from the beginning to be independent from the installed version of Java. You write C++ code for JNI, compile it and link it without reference to any library (there is no "jni library"). The same compiled C++ JNI code can be used from any implementation or version of the JVM (at least a "recent" version). Python, on the other hand, adopted a very unfortunate model where native code should be compiled using a Python command (not a classical gcc command) and the result is only usable with the same version of Python. If I compile my native code in a Python 3.9 environment, the binary may not be usable if the target system uses Python 3.8. Since this is not acceptable for project distribution, I had to use the ugly "ctype" Python interface. Developing the bindings is much more complex and dangerous in terms of bindings but the native binary can be used from any version of Python on the target system.

So, my question about Javascript binding is: how does it look like? A Java-like model or a Python-like model? Is the native binary usable from any (not too obsolete) version of nodejs in the target system?

lars18th:
It's a pain that NodeJS support will not be completed finally. Could the original author (@jkuemmele) please comment?

lelegard:
Closed after one year without update.

Issue #961, opened 2022-03-17
[Q] DVB-T2 support for dektec modulators


macczy:
Hi,
I have a dektec DTU-315 modulator, which I wanted to use with ts duck to stream DVB-T2 streams. At first started with simple command:

~~~
./tsp -v -I file /path/to/test_stream.ts --infinite -O dektec --modulation DVB-T2 --frequency 642000000
* file: initial input bitrate is 24,128,572 b/s
* dektec: using Dektec device 0, output channel 0 (DTU-315 port 1)
* dektec: setting output TS bitrate to 39,160,146 b/s
* dektec: setting output carrier frequency to 642,000,000 Hz
* dektec: output fifo size: 16,777,216 bytes, max: 16,777,216 bytes, typical: 16,777,216 bytes
* dektec: initial output bitrate: 39,160,146 b/s
* dektec: Will preload FIFO before starting transmission. Preload FIFO size: 13,421,696 bytes.
* dektec: DTU-315 output FIFO load is 13,421,696 bytes, starting transmission
~~~

Then on TV you can tune to, for example, ITV channel.
Unfortunately, the output is not as expected. The video is sped up and, well let's say it's constantly jumping. Please take a look at the attached recording, as it should be clear from it what I mean: 
https://drive.google.com/file/d/1IUh2121fSmAJQhBWKLAddqxR66riA7IK/view?usp=sharing

The stream I've used is a recording from UK. The file was quite big so I've cut a small piece of it. It's enough to reproduce, so I'm also attaching this file: https://drive.google.com/file/d/1BKrarpZYXmeQ9-lgqJ8czVNY58NTypkf/view?usp=sharing

I've tried changing bitrate manually, but that does not seem to do the job, so I wonder if you have any suggestions how we could stream it properly.

For reference - the same stream on same device works properly when using DekTec's StreamXpress instead of ts duck, so I'm positive the problem does not lay in the stream itself, but rather in how we try to use TSduck or some bug in the implementation.

#### Environment:
 - OS: Linux
 - OS version: Ubuntu 20.04
 - TSDuck full version: 3.31-2711
 - Installation type: building from source

lelegard:
Hello @macczy 

We can see this:
~~~
* file: initial input bitrate is 24,128,572 b/s
...
* dektec: setting output TS bitrate to 39,160,146 b/s
~~~

[...]

PR #970, opened 2022-04-06
Handle "input-modulation" flag for DVB-T2 dektec output

macczy:
"--input-modulation" flag allowed to automatically calculate modulation
parameters when streaming with dektec output plugin from file using DVB-T modulation.
However, it was ignored for DVB-T2. Therefore handle it in a similar manner for DVB-T2.

#### Affected components:
dektec output plugin

Recently I've asked about related problems here: https://github.com/tsduck/tsduck/issues/961
I was advised to use additional stuffing, and that worked for a while, but, in afterthought, didn't seem like long term solution. Then I discovered, that I have similar issues when streaming in DVB-T. But those could be easily solved with "input-modulation" flag. Therefore I would like to add support for this flag for DVB-T2 as well.


lelegard:
Thanks @macczy.
I will review this. In the meantime, please replace my name with your full name in new source files you create. You are the author, I am not.

macczy:
Hi @lelegard,
I've updated the name in a new commit for now. I assume we can just squash them during merge if you accept my change.

macczy:
Looking at CI errors, I had to move new files to the dektec-specific directory, so that's done.
There was also this:
`XXXX  1 tests FAILED, 1614 tests passed, review failed tests in /home/runner/work/tsduck/tsduck-test/tmp/.failed
make: *** [Makefile:77: test-suite] Error 1
==== tmp/test-019.tsp.output.dektec.linux.help.diff
195,196c195,196
<       Otherwise, if the specified modulation is DVB-T, try to guess some
<       modulation parameters from the bitrate.
---
>       Otherwise, if the specified modulation is DVB-T or DVB-T2, try to guess
>       some modulation parameters from the bitrate.
Error: Process completed with exit code 2.`
I had to change the help message a bit, so I assume that's something to update in CI? Or is there some testing file I couldn't find that holds the old help messages?

lelegard:
> Or is there some testing file I couldn't find that holds the old help messages?

The test suite also checks the help text of all commands and plugins to detect potential breaks. When options are added or modified, the reference texts must be manually regenerated and checked. There is a procedure for that. I will handle that, don't worry about that "failure".

If you are interested in the way TSDuck is tested, you may have a look [here](https://tsduck.io/doxy/testing.html) and [here](https://github.com/tsduck/tsduck-test).

PR #1050, opened 2022-08-30
Allow Headers To Be Used From C++20 Code

mich181189:
When using TSDuck from a C++20 application, the headers throw various
errors. This commit tidies them up.

The library itself still does not build cleanly as C++20 (biggest problem is the
initialisation of the const TSPacket instances in tsTSPacket.c) however this is enough for 

#### Affected components:
Various types in libtsduck

Most changes are removing the template arguments from destructors. These redundant arguments cause warnings (which frequently become errors) when building a C++20 target

There's also a change to `operator==(NUM_T x1, const ts::FloatingPoint<FLOAT_T,PREC>& x2)`  because C++20 seems to reorder operator arguments, which causes the original implementation of this to become in infinite loop (again, this caused a compiler warning which became an error) - the fix is to explicitly call the intended operator, preventing the infinite loop

### DIFF
--- a/src/libtsduck/base/algo/tsMessageQueueTemplate.h
+++ b/src/libtsduck/base/algo/tsMessageQueueTemplate.h
@@ -46,7 +46,7 @@ ts::MessageQueue<MSG, MUTEX>::MessageQueue(size_t maxMessages) :
 }
 
 template <typename MSG, class MUTEX>
-ts::MessageQueue<MSG, MUTEX>::~MessageQueue<MSG, MUTEX>()
+ts::MessageQueue<MSG, MUTEX>::~MessageQueue()
 {
 }
 
--- a/src/libtsduck/base/report/tsReportFileTemplate.h
+++ b/src/libtsduck/base/report/tsReportFileTemplate.h
@@ -71,7 +71,7 @@ ts::ReportFile<MUTEX>::ReportFile(std::ostream& stream, int max_severity) :
 //----------------------------------------------------------------------------
 
 template <class MUTEX>
-ts::ReportFile<MUTEX>::~ReportFile<MUTEX>()
+ts::ReportFile<MUTEX>::~ReportFile()
 {
     GuardMutex lock(_mutex);
 
--- a/src/libtsduck/base/system/tsResidentBufferTemplate.h
+++ b/src/libtsduck/base/system/tsResidentBufferTemplate.h
@@ -126,7 +126,7 @@ ts::ResidentBuffer<T>::ResidentBuffer(size_t elem_count) :
 //----------------------------------------------------------------------------
 
 template <typename T>
-ts::ResidentBuffer<T>::~ResidentBuffer<T>()
+ts::ResidentBuffer<T>::~ResidentBuffer()
 {
     // Unlock from physical memory
     if (_is_locked) {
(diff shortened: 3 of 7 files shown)


lelegard:
Hi @mich181189 

Could you please specify which compiler and version you used in C++20 mode?

I tried to compile the current version using `make CXXFLAGS_STANDARD=-std=c++20` and the result depends on the compiler (on a Fedora 36 system):
- With gcc 12.2.1, we get errors on the destructor templates and on the ambiguous / reversed operator `==`.
- With clang 14.0.5, we only get errors on the ambiguous / reversed operator `==` only.

But there is worse. With your fix and the current default of C++11, the code compiles with gcc and msvc but fails with clang 13.0 on macOS. This can be seen [in the PR checks logs](https://github.com/tsduck/tsduck/runs/8099056625):
~~~ 
/Users/runner/work/tsduck/tsduck/src/libtsduck/base/types/tsSafePtrTemplate.h:38:21: error: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Werror,-Wdtor-name]
759
ts::SafePtr<T,MUTEX>::~SafePtr()
~~~

Note, however, that the [PR check for clang on Linux](https://github.com/tsduck/tsduck/runs/8099055576), using clang 10.0 seems to pass.

So, this is not as simple as it seems. There are conflicting requirements between the various compilers, even the versions of a given compiler, and the requested level of standard. This is not the fist time I see this but this needs to be checked and a more robust fix needs to be found to be compatible with all versions of all compilers using all levels of standards.

TSDuck is still used by people using old compilers. At some point, I decided to drop support for pre-C++11 but we still need to support a wide range of compilers and versions.

I will look into this one.

mich181189:
I'm using Fedora 35 - GCC 11.3.1 (I have some other stuff that doesn't yet build on 36 and I haven't had chance to fully fix it yet)

I'm happy to use my branch until you figure this out - I realise it's not necessarily a simple problem! C++20 does seem to have introduced some slightly problematic changes for some code

~~~
$ gcc -v
Using built-in specs.
COLLECT_GCC=/usr/bin/gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/11/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,objc,obj-c++,ada,go,d,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --with-linker-hash-style=gnu --enable-plugin --enable-initfini-array --with-isl=/builddir/build/BUILD/gcc-11.3.1-20220421/obj-x86_64-redhat-linux/isl-install --enable-offload-targets=nvptx-none --without-cuda-driver --enable-gnu-indirect-function --enable-cet --with-tune=generic --with-arch_32=i686 --build=x86_64-redhat-linux --with-build-config=bootstrap-lto --enable-link-serialization=1
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.3.1 20220421 (Red Hat 11.3.1-2) (GCC) 

~~~

lelegard:
Hi @mich181189 

I updated all code in TSDuck to be compatible with C++20. This required quite a number of modifications. Especially, starting with C++20, you must no longer declare `operator!=`  when the corresponding `operator==` has been declared. C++20 automatically uses the negation of == when != is used. Since a custom `operator!=`  is always declared with ` const`  attributes, any != test with at least one non-const operand results in an "ambiguous operator" error. The custom `operator!=`  is no longer a direct match and there are two equally ambiguous candidates.

We may anticipate additional errors with other compilers or other versions of compilers in C++20 mode. It may take some time to reach a stable state.

lelegard:
I have updated the CI workflow to build TSDuck using C++14 and C++20 standards on push and PR to track potential future incompatibilities.

See an example run here: https://github.com/tsduck/tsduck/actions/runs/2973393337

PR #1062, opened 2022-09-29
add metadata for next generation video codecs (VVC, AVS3 & EVC) according tp latest industry standards

paulhiggs:
…ding to ITU-T H.222.0 (2021-06) and DVB drafts of ETSI EN 300 468 and ETSI TS 101 154


lelegard:
Thanks for the update.

PR #1071, opened 2022-10-10
Update attribute cardinality and evaluatyion 

paulhiggs:
HDR_WCG_Idc and video_properties_tag attributes are always present in the descriptor.
semantic definition of video_properties_tag depends on HDR_WCG_idc to select colour format (BT709 vs BT2020 vs BT2100)


lelegard:
Hi @paulhiggs 

Did you try the display routine? It seems to me that the following won't work:
~~~
    const uint8_t vprop_tag = buf.getBits<uint8_t>(4);
    disp << ", video properties: " << VVCVideoDescriptor::VVCVideoProperties(hdr_wcg_idc, vprop_tag)
         << " (" << vprop_tag << ")" << std::endl;
~~~

Specifically, `<< vprop_tag << ` will probably display some garbage. When `vprop_tag` was declared as `uint16_t`, the value was interpreted as integer and displayed in decimal. With the updated declaration as `uint8_t`, it will be interpreted as a character, ie. the ASCII code of a character, and displayed as that character. This is the old and annoying inability of C/C++ to discriminate between characters and 8-bit integers.

lelegard:
> UStream::Format(u"%d", {vprop_tag}) call which I assume will display it correctly as a decimal.

Yes it will. But `<< int(vprop_tag) << ` is probably simpler.

PR #1072, opened 2022-10-10
Update decimal display of descriptor values

paulhiggs:
Cast byte values as int o they display in decimal not ASCII


lelegard:
Already done in parallel !

PR #1074, opened 2022-10-11
add TTML Subtitling Descriptor per ETSI EN 303 560

paulhiggs:
#### Affected components:
Additional descriptor in the TSDuck library

Adds the TTML_subtitling_descriptor according to [ETSI EN 303 560 v1.1.1](https://www.etsi.org/deliver/etsi_en/303500_303599/303560/01.01.01_20/en_303560v010101a.pdf). Verified against the published [DVB TTML test sequences](https://dvb.org/specifications/verification-validation/dvb-ttml-subtitling-test-streams/).

### DIFF
--- a/src/libtsduck/dtv/descriptors/tsMPEG4TextDescriptor.cpp
+++ b/src/libtsduck/dtv/descriptors/tsMPEG4TextDescriptor.cpp
@@ -263,7 +263,9 @@ void ts::MPEG4TextDescriptor::DisplayDescriptor(TablesDisplay& disp, PSIBuffer& 
                 disp << ", length: " << textConfigLength << std::endl;
                 if (textConfigLength > 0) {
                     if (textFormat == 0x01) {
-                        disp << margin << TimedText_TS26245(buf.getBytes(textConfigLength));
+                        // Need to extract string first because of issue with MSVC in C++20 mode.
+                        const UString line(TimedText_TS26245(buf.getBytes(textConfigLength)));
+                        disp << margin << line;
                     }
                     else {
                         disp << margin << UString::Dump(buf.getBytes(textConfigLength), UString::SINGLE_LINE);
--- a/src/libtsduck/tsVersion.h
+++ b/src/libtsduck/tsVersion.h
@@ -44,4 +44,4 @@
 //!
 //! TSDuck commit number (automatically updated by Git hooks).
 //!
-#define TS_COMMIT 2929
+#define TS_COMMIT 2930


paulhiggs:
@lelegard Can you update the private data specifier used in ````reference/test-115.pack.ts```` to ````0x41565356````. This is the value that is being registered through DVB Services for AVS.

lelegard:
> Can you update the private data specifier used in reference/test-115.pack.ts to 0x41565356. This is the value that is being registered through DVB Services for AVS.

Could you be more specific? Because the problem may be slightly different from what you think.

The private data specifier I recently added in TSDuck was a fake, just like ATSC or ISDB. It is an internal hack inside TSDuck due its legacy design where, internally, a descriptor with a tag > 0x80 shall be associated with a private data specifier. But these fake private data specifiers are never used in a private data specifier descriptor. So, in test-115, there is no private data specifier descriptor.

Do you mean that, following our previous discussion, you recently obtained from DVB an official private data specifier for AVS? In that case, we need to requalify this value as a real private data specifier and no longer a fake one. This also mean that the test-115 must be updated with a private data specifier descriptor, before the AVS descriptor. This would be a nice solution since this is now DVB-compliant.

I do not see this value yet in https://www.dvbservices.com/identifiers/private_data_spec_id but we may have to wait a little bit more to have the list updated.

I can take care of all the implications in the code.

Could you please just confirm that this interpretation in correct? Is there any other private descriptor to be defined under this private data specifier?

paulhiggs:
> Could you be more specific? Because the problem may be slightly different from what you think.
At some stage, you defined ````PDS_AVS = 0x41565333, //!< Fake private data specifier for AVS descriptors (value is "AVS3" in ASCII).```` in ````tsPSI.h```` and I presume created the test bitstream ````reference/test-115.pack.ts```` to include a private_data_specifier_descriptor with this value along with an ````[AVS3_video_descriptor](https://github.com/tsduck/tsduck/blob/master/src/libtsduck/dtv/descriptors/tsAVCVideoDescriptor.cpp)````. This presumption is based on the fact that the test suite CI passed when this value was used, i.e.
````
    - Descriptor 0: AVS3 Video (T/AI 109.6) (0xD1, 209), 8 bytes
      Profile ID: High-10 (0x32), Level ID: 8.4.60 (0x55), Multiple frame rate: false
      Frame rate code: 60 (8), Sample precision: 10-bit (2), Chroma format: 4:2:0 (1)
      Temporal ID: false, TD mode: true
      Library stream: false, Library picture enable: true
      Colour primaries: 9, Transfer characteristics: 14, Matrix coefficients: 8
````
Although I may have misunderstood how this was occurring.

Now the AVS Working Group of China is in the process of obtaining the value ````0x41565356```` from DVB Services, so I committed a change to ````tsPSI.h```` accordingly - now the test suite does not run "as expected"

> Could you please just confirm that this interpretation in correct? Is there any other private descriptor to be defined under this private data specifier?
There are no other private descriptors for AVS3 at this time (there could be others when additional/advanced profiles of AVS3 are released)

lelegard:
>  This presumption is based on the fact that the test suite CI passed when this value was used, i.e.

In fact not. The "fake" PDS are an internal hack of TSDuck (and I am ashamed of it). Internally, a descriptor with tag > 0x80 cannot exist without an associated PDS. This is due to the original "DVB only" design of TSDuck. When ATSC and ISDB were introduced, the fastest way to integrate those descriptors was to introduce a fake internal PDS.

But these fake PDS are internal only. They are never output in descriptor lists since the corresponding private data specifier descriptor is a DVB one which would be illegal in the context of ATSC or ISDB.

I have question about the AVS3 descriptor. Why didn't you try to have it integrated in ISO 13818-1, just like an HEVC or VVC video descriptor? It would use another MPEG-assigned tag, probably an "extended descriptor structure". However, it would be automatically valid in all contexts, DVB, ATSC, ISDB, etc. Currently, you define a way to introduce it in DVB stream using an official DVB-assigned private data specifier for you. But how will it be integrated in non-DVB streams? ATSC, ISDB? IP streaming? Did you get some approval or official assignment from the various standard bodies?

paulhiggs:
Thanks for the excellent clarification - I hope to have a registered PDS Id this week.

For now I will revert back to 'AVS3' equivelant so that the test suite does not fail my current commits!

> I have question about the AVS3 descriptor. Why didn't you try to have it integrated in ISO 13818-1, just like an HEVC or VVC video descriptor? It would use another MPEG-assigned tag, probably an "extended descriptor structure". However, it would be automatically valid in all contexts, DVB, ATSC, ISDB, etc. Currently, you define a way to introduce it in DVB stream using an official DVB-assigned private data specifier for you. But how will it be integrated in non-DVB streams? ATSC, ISDB? IP streaming? Did you get some approval or official assignment from the various standard bodies?

Well, the answer is quite straightforward and not one I have control over - ISO, or perhaps specifically JTC 1/SC 29, won't include any technology that they do not develop. At present we are only working with DVB, if ATSC or ISDB want to use it for broadcast we will work with them to identify their preferred mechanism.
For IP Streaming we primarily use DVB-DASH or HLS and these mechanisms have a different header format - I believe a large part of the actual AVS3 video sequence header is in the ISO BMFF header bits. That should all be visible in the T/AI 109.6 specification.

lelegard:
> For IP Streaming we primarily use DVB-DASH or HLS and these mechanisms have a different header format - I believe a large part of the actual AVS3 video sequence header is in the ISO BMFF header bits. That should all be visible in the T/AI 109.6 specification.

HLS also supports the TS format. There is an input/output `hls` plugin in TSDuck. I believe that DASH also supports the TS format but it seems less common.

paulhiggs:
> HLS also supports the TS format. There is an input/output `hls` plugin in TSDuck. I believe that DASH also supports the TS format but it seems less common.
Yes, but we can probably signal codec media capabilities in the associated manifest and not rely on the transport stream..

paulhiggs:
> And more specifically, which standard shall we assign to the AVS3 descriptor?

The descriptor is defined by **AVS** and published in T/AI 109.6 and referenced by **DVB** in A001 and  **ETSI** in TS 101 154. Ideally AVS 'owns' the standard that defines the descriptor, hence I did add them as ````ts::Standards::AVS````. 
We also know that [AOM](https://aomedia.org) is defining an AV1_video_descriptor() in the [AV1 Specification for carriage inside MPEG2-TS](https://aomediacodec.github.io/av1-mpeg2-ts/) so we may need to eventually add support for them.
> 
> AVS is a codec working group. The "standards" which are referenced and managed by TSDuck are "transport stream layer standards", starting with MPEG, the common root, then the various branches DVB, ATSC, ISDB and subsequent subdivisions of ISDB which is a loosely defined standard with anarchic regional variants.

So AVS3 video coding carried in a DVB transport stream should be valid - see the latest DVB [Bluebook A001](https://dvb.org/wp-content/uploads/2022/02/A001r20_Use-of-Video-and-Audio-Coding-in-Broadcast-and-Broadband-Applications_Draft_TS_101-154-v281_July-2022.pdf). DVB always prefers to 'reference when  it can and standardize when it cannot'
> 
> Specific types of contents, including video codecs but not only, have either a top-level signalisation, defined in MPEG ISO 13818-1, or a declination of variants in each standard (I mentioned the example of AC-3).

I need to look at how AC-3 is signalled in an MPEG transport stream, but there are many ways (unfortunately)  to support signalling of non-MPEG codecs.
> 
> Since you obtained (or are about to obtain) an official DVB private data specifier, the AVS3 descriptor becomes a valid DVB private descriptor. It consequently bring several points:
> 
> * AVS should be removed as a standard in TSDuck.
But AVS created the descriptor for use in any and all regional transport stream variants. If AVS3 is adopted in ISDB then the T/AI 109.6 would still be references and the descriptor still owned and maintained by AVS.

> * The AVS3 descriptor becomes a DVB descriptor.
I don't see that happening in the real world - DVB will reference anyone's descriptor but not take ownership of it.

> * What about non-DVB environments?
As above..

-- the world is a-changing --

paulhiggs:
@lelegard 
Do you know why this error is occurring in the CI of my pull request?
image

I have not changed anything with the Dektec functionality.

lelegard:
> Do you know why this error is occurring in the CI of my pull request?

Yes, it's my fault. In the process of making .xml and .names files modular, if made the generation of the file `tsduck.dektec.names` dynamic at build time (instead of regenerating from time to time and store it as a source file).

The generation is done by a python script. On Windows, it is triggered as an "AfterBuild" target. It works fine on my Windows VM's. However, on the CI Windows runner, the generated file is not created, or disappears, without error message. I am investing this.

Do not worry about this errors.

lelegard:
I have fixed two Windows integration issues with the GitHub Windows CI runner. First, you need to specify "python" when running  a python script in a GitHub CI runner while using the script name was sufficient elsewhere. The second issue is due to the deployment of a new version of Visual Studio introducing a different interpretation in the "Exclude" directive in the project files.

If you integrate your patches with the master branch, the tests should pass.

lelegard:
Hi @paulhiggs 

I am currently integrating this. I will fix a few mishaps.

I will also remove AVS as a standard (meaning not a "TS standard" such as DVB or ATCS) because the AVS3 descriptor now becomes a valid DVB private descriptor.

lelegard:
Hi @paulhiggs 

I have merged all your updates, up to e6c8ff9914f30582ceae9b68c56d3097f715ed9b (fix fractional 30Hz frame rate name). The rest seems a cross-merging issue. Please confirm that everything is here. You may close this PR if this is the case.

I had to fix a few things in the MPEG4 text descriptor after adding test patterns in test-115.

I plan to write a complete tutorial, in the [programmer's guide](https://tsduck.io/doxy/index.html), about adding new tables and descriptors, including full test plan. This would probably help.

paulhiggs:
Thanks @lelegard 
I will certainly look more at the programmers guide and understand the test environment going forward.

PR #1080, opened 2022-10-17
Additional descriptors from the TODO list

paulhiggs:
Added MuxCode descriptor and FmxBufferSize descriptor.
Note that ISO/IEC 13818-1 refers to 14496-1 for FmxBufferSize and while this was present it early editions it is not currently the case (this has been raised to the editors of both specifications in MPEG)


lelegard:
Hi Paul

Thanks a lot for taking the lead on the missing descriptors. I implemented so many of them that I admit I gave up the TODO list a while ago.

By experience, there are many potential little issues when implementing a descriptor or table. It is necessary to add a test for every descriptor. I recommend the following scenario:

Checkout repositories `tsduck` and `tsduck-test` side by side in the same parent directory.

Implement a new test or supplement an existing test. In your case, test-115 would be your playground. For each newly implemented descriptor, add a few samples in `input/test-115.xml`. One may be enough. One or two more can be useful to test with and without optional parts for instance. Don't care about the consistency of the values in the real-world, the point is to stress the code, exhibit potential bugs. For instance, use distinct values, ints or bools, for adjacent binary fields, just to make sure the right values are extracted from the right place. Again, no matter if the values are inconsistent in terms of MPEG or DVB semantics.

Compile TSDuck in its workspace. If you use Windows, prefer a 64-bit build, release or debug.

Run the test to regenerate the reference files for the new descriptors. In a git bash window, go to directory `tsduck-test` and run `tests/test-115.sh --dev --init`. The option `--dev` means use the compiled binaries in the neighbour `tsduck` work area, release version. Use `--debug` instead of `--dev` if you compiled the debug version.

Normally, no message should appear. In case of error message, an early stage of the test failed and the errors are generated by the subsequent phases. Review the errors in the files `reference/test-115.*.log`.

If no message appear, then manually review the initial production of the test in files `reference/test-115.*`. Since you supplement an existing test, `git diff` is the best way to review the additions. With the "standard SI test scenario" in test-115, focus on the output .xml, .txt, .log files. The .log files should be all empty (= no error). The XML files should contain the same thing as your input file (apart from reformatting and explicit default values). The .txt files are produced by the `DisplayDescriptor` static method. Review the .xml and .txt files carefully. If they do not match the input or your expectation, there is a problem, either in serialisation (input XML to C++ to binary) or deserialisation (binary to C++ to output XML) and it is time to review the code or debug.

The .json files are automatically built from the .xml files and there are few to no reason they are incorrect. The .ts and .bin files are in the middle of serialisation / deserialisation and they are probably correct if the output is correct. Nonetheless, it is still possible that a "bug-compatible" error exists in serialisation and deserialisation, such as the one you found on inverted boolean logic in HEVC and VVC video descriptor. So, you may check the binary if you can and have time.

Once the reference files are OK, add them in the repo. If you implement several descriptors, I recommend to test one first. Then, when the reference files are OK, use `git add .` before testing the next one. Thus, `git diff` always reports the differences for the currently tested descriptor, one by one.

Once everything is OK, you may run `tests/test-115.sh --dev` (without `--init`). It runs the test again but generates the files in a temporary area and compares them with the reference files. This is the non-regression which will be part of every push. No error message should appear. Otherwise, you have a non-reproduceable test output (very unlikely in the case of a standard SI test).

Finally, push the two repositories and issue a PR for each repo, `tsduck` and `tsduck-test`.

At that time, you can be confident with your code.

You may find additional explanations [here](https://github.com/tsduck/tsduck-test) and [here](https://tsduck.io/doxy/testing.html). This reponse was customised to your test.

paulhiggs:
Thanks, I will delve more into this.
After setting up ````tsduck-test```` alongside ````tsduck```` and building the later, I get the following from by bash shell with trying to init ````test0115````
image

Not sure why the host name is being appended by ````test-115.sh````

paulhiggs:
OK, so I updated ````input/test-115.xml```` with some sample descriptors but there are some issues with using ````tests/test-115.sh --debug --init````
````
test-115
sed: can't read /z/development/tsduck-test/reference/test-115.decompile.xml: No such file or directory
sed: can't read /z/development/tsduck-test/reference/test-115.decompile.xml: No such file or directory
sed: can't read /z/development/tsduck-test/reference/test-115.decompile.strict.xml: No such file or directory
sed: can't read /z/development/tsduck-test/reference/test-115.decompile.strict.xml: No such file or directory
sed: can't read /z/development/tsduck-test/reference/test-115.decompile.json: No such file or directory
sed: can't read /z/development/tsduck-test/reference/test-115.decompile.json: No such file or directory
````

Then on a run without ````--init````
````
$ tests/test-115.sh --debug
test-115
XXXX  test-115: Reference output /z/development/tsduck-test/reference/test-115.compile.bin missing
----  test-115: Test test-115.compile.log passed
sed: can't read /z/development/tsduck-test/tmp/test-115.decompile.xml: No such file or directory
sed: can't read /z/development/tsduck-test/tmp/test-115.decompile.xml: No such file or directory
XXXX  test-115: Reference output /z/development/tsduck-test/reference/test-115.decompile.xml missing
----  test-115: Test test-115.decompile.log passed
sed: can't read /z/development/tsduck-test/tmp/test-115.decompile.strict.xml: No such file or directory
sed: can't read /z/development/tsduck-test/tmp/test-115.decompile.strict.xml: No such file or directory
XXXX  test-115: Reference output /z/development/tsduck-test/reference/test-115.decompile.strict.xml missing
----  test-115: Test test-115.decompile.strict.log passed
sed: can't read /z/development/tsduck-test/tmp/test-115.decompile.json: No such file or directory
sed: can't read /z/development/tsduck-test/tmp/test-115.decompile.json: No such file or directory
XXXX  test-115: Reference output /z/development/tsduck-test/reference/test-115.decompile.json missing
----  test-115: Test test-115.decompile.json.log passed
XXXX  test-115: Reference output /z/development/tsduck-test/reference/test-115.compile.json.bin missing
----  test-115: Test test-115.compile.json.log passed
XXXX  test-115: Test test-115.cmp.json.log FAILED, check /z/development/tsduck-test/tmp/test-115.cmp.json.log.diff
----  test-115: Test test-115.pack.ts passed
----  test-115: Test test-115.pack.log passed
----  test-115: Test test-115.tstables.txt passed
----  test-115: Test test-115.tstables.xml passed
----  test-115: Test test-115.tstables.log passed
----  test-115: Test test-115.tstables.strict.xml passed
----  test-115: Test test-115.tstables.strict.log passed
----  test-115: Test test-115.tstabdump.txt passed
----  test-115: Test test-115.tstabdump.log passed
````
the tests seem to pass but there are other issues reported.

paulhiggs:
> Finally, push the two repositories and issue a PR for each repo, ````tsduck```` and ````tsduck-test````.

Do I just put the updated ````input/test-115.xml```` into the PR or all the ````reference/test-115.*```` files too?

lelegard:
> Do I just put the updated input/test-115.xml into the PR or all the reference/test-115.* files too?

Push everything that `git add .` adds, including the reference files. But I see that you have already done it correctly.

The idea of the test suite is that you generate reference files each time you add or update the tests, manually verify the correctness of the reference files once, and then keep those reference files in the repo to verify the non-regression tests which will run thousands of time in the future. This is how we can feel confident that changing a tiny little thing somewhere in the future will not break something else in the code without being immediately detected.

lelegard:
I see that you currently add some commits. Please add a note here when you think the code is complete and can be merged.

PR #1084, opened 2022-10-19
decoding of HEVC operation point descriptor. 

paulhiggs:
this version treats the HEVC Profile Tier Level information (the sets of 96 bits) opaquely - a future update will dissect these


lelegard:
I would recommend `for (size_t j = 0; j < _ES.size(); j++)` because you iterate over an array index and an index is always of type `size_t` in C++. In the two loops. The alert is perfectly valid. If a malicious user injects an XML text with 256 or more subsections, your code will loop forever with `uint8_t` because it loops back from 255 to 0 and never reaches `size()`.

You may say that this would be an invalid table and you are right. But the art of hacking is to find the right invalid input which fools code which does not correctly check its input to make it do something dangerous or damaging (here, a denial of service).

I am pleasantly surprised by the quality of GitHub code scanning. This is part of their effort to improve the security of open source code. And it seems to work!

Issue #1090, opened 2022-10-25
[FEAT] Unicable/SCIF support


abainbridge:
#### Is your feature request related to a problem?
My building uses unicable to distribute satellite signals.

#### Proposed solution:
Add support for unicable.

#### Alternatives:
Use a different program or rewire all the world's buildings that use unicable.

#### Additional information:
I've already implemented a working prototype for this. Before I continue, I thought I'd see whether a pull request that implements this feature is something that would interest the project. Also, it would be good to discuss some design details.

#### Command line parameters

Currently, to tune to DiSEqC I might use a command like:
`tsp -I dvb --adapter 0 --delivery-system DVB-S2 --frequency 985000000 --fec-inner 3/4 --polarity vertical --satellite-number 3  --symbol-rate 29950000 -O file`

All of these parameters continue to be needed to tune via Unicable and there are a whole group of extra parameters needed. Adding a new DVB option like the following might be best:
```
--unicable=<version>,<userband slot>,<userband frequency>[,<pin>]
    version: Which Unicable version to use. Must be 1 or 2.
    userband slot: Which userband slot to tell the Unicable switch to output your channel on. For version 1, must be in the range 1 to 4. For version 2, lmust be in the range 1 to 32.
    userband frequency: Frequency the local tuner must tune to in order to receive the specified userband. In MHz.
    pin: Version 2 only. PIN used to protect the userband for installation aid purposes.
```
The man page text for other parameters might need changing too.

`---frequency` was:
```
      Carrier frequency in Hz (all tuners). There is no default.
```
Might become:
```
      Frequency at which the channel is broadcast, in Hz (all tuners). There is
      no default. In the case of DiSEqC and Unicable, the actual frequency the
      tuner tunes to is calculated from this, to take account of hetrodyning.
```

`--satellite-number` was:
```
      Used for satellite tuners only. Satellite/dish number. Must be 0 to 3
      with DiSEqC switches and 0 to 1 fornon-DiSEqC switches. The default is 0.
```
Might become:
```
      Used for satellite tuners only. Satellite/dish number.
      For DiSEqC 1.0, must be 0 to 3.
      For Unicable 1, must be 0 to 1.
      For Unicable 2, must be 0 to 63.
      The default is 0.
```

lelegard:
Hi @abainbridge 

Thanks for the contribution. Do you have public reference to Unicable specification and how it is interfaced on Linux and Windows?

[...]

PR #1101, opened 2022-11-03
Minimal hacky changes needed to support satellite tuning via a Unicab…

abainbridge:
…le switch.

#### Related issue (if any):
[]
https://github.com/tsduck/tsduck/issues/1090
(Sorry, I don't know what the #nnn notation is).

#### Affected components:
dvb plugin.
C++ component libtsduck/dtv.

Add ts::TunerDevice::configUnicableSwitch() and call it instead of ts::TunerDevice::dishControl() in ts::TunerDevice::tune().


lelegard:
Not generic enough for integration.

Moreover, this PR is based on a state of the repo which is no longer in its history (see #1490) and it cannot be merged.

PR #1107, opened 2022-11-08
Latest updates to descriptors

paulhiggs:
Check reserved bits are set according to the relevant standard
Update and improve output formatting
Include a common routine to tabulate hexadecimal values (tsTabluateVector)


lelegard:
Hi Paul,

Just one question about the `TabulateVector` functions. Is there a reason why you did not use one of the `UString::Dump()` static methods? They are quite versatile and used everywhere hexa/ascii/binary dump is required.

Moreover, in the `DisplayDescriptor` methods, you can also use specialised methods from the `TablesDisplay` instance: `displayPrivateData()` or `displayExtraData()`. They call `UString::Dump()` to format the dump but they add some encapsulation for presenting descriptor or section data.

Before merging this one, I would like to make sure that it does not introduce redundant code in the main library.

paulhiggs:
I did not look deeply into the `UString::Dump()` mostly because I knew what I needed to to and that it was already done that way in several of my descriptors. In fact `TabulateVector` mostly just takes the existing code and moves it, but it does also add the `vector<boolean>` option.
I had seen `UString::Dump` being used for private data and thought that I did not need all the line and byte numbering provided there.
For me, `tsTabulateVector()` seems to be more useful for a set of values after the label, i.e....
````
      LCEVC stream tag:  01 23 45 17 18 19 20 21
                         22 23 24 25 25 26 27 28
                         28 30
````
I could look into using `UString::Dump()` if you prefer!

lelegard:
> I could look into using UString::Dump() if you prefer!

Also have a look at `TablesDisplay::displayPrivateData()`. It is used in many descriptors to display stuff similar to the "LCEVC stream tag" you mention.

For the sake of code consistency, I would prefer to keep a common set of tools to handle the signalisation.

Let me know what is missing, such as `vector<boolean>`, and I will add them.

lelegard:
Or, alternatively, since your new functions are bound to `TablesDisplay`, I would prefer to move them as methods of `TablesDisplay`, under the name `displayVector()`, as specialised companions to `displayPrivateData()`. This seems more consistent than having independent functions.

This is maybe the most appropriate compromise.

paulhiggs:
I am happy to change from `TabulateVector()` ..
````
    - Descriptor 2: MPEG-2 Extension (0x3F, 63), 46 bytes
      MPEG extended descriptor: LCEVC linkage (0x18, 24)
      LCEVC stream tag:  01 23 45 17 18 19 20 21
                         22 23 24 25 25 26 27 28
                         28 30 41 42 43 45 46 46
                         46 47 48 49 45 51 52 53
                         54 54 56 45 45 99 98 97
                         96 95 94 93
````
to `UString::Dump()` ..
````
    - Descriptor 2: MPEG-2 Extension (0x3F, 63), 46 bytes
      MPEG extended descriptor: LCEVC linkage (0x18, 24)
      LCEVC stream tag: 01 23 45 17 18 19 20 21 22 23 24 25 25 26 27 28 28 30 41 42 43 45 46 46 46 47
48 49 45 51 52 53 54 54 56 45 45 99 98 97 96 95 94 93
````

For the boolean output, I was looking for something like this....
````
channel map: X..X...X.XXX.XX...XX
             ....X..XX...X..X..X.
             XXX...X.XXX..X..XX.X
````

paulhiggs:
> Or, alternatively, since your new functions are bound to `TablesDisplay`, I would prefer to move them as methods of `TablesDisplay`, under the name `displayVector()`, as specialised companions to `displayPrivateData()`. This seems more consistent than having independent functions.
That would be a good solution

lelegard:
> That would be a good solution

OK, we have a deal :) I let you update the code and the PR accordingly.

Issue #1125, opened 2022-11-23
[FEAT] HLS output "--no-align"


lars18th:
#### Is your feature request related to a problem?
I want to produce HLS segment files that will be a 1:to:1 representation of a CBR stream. The HLS output plugin can write to files, and the HLS input plugin can read from a file URI. So we can produce these segments to process them off-line. However, the current HLS output plugin is inserting a copy of PAT and PMT tables at the start of each segment:

[ Documentation "hls (output)" section ]
_**-a | --align-first-segment**
Note that all subsequent output segments always start with a copy of the last PAT and PMT,
on a video PES packet boundary, with or without this option._

#### Proposed solution:
Add to the `hls (output)` plugin a new option `--no-align` that will produce *clean* segments. These segments are a bit perfect representation of the TS packets, without any reordering or filtering. The generated segments could them be joined (`cat foo-000000.ts foo-000001.ts foo-000003.ts > stream-foo.ts`) to reconstruct the original stream. Or they can be used with the HLS input plugin to read them using the standard playlist `foo.m3u8`.

As an addition, it will be useful to provide an alternative mode like `--no-align-extended`. In this case the segments are the same, but for each of them another file with the missing starting PAT and PMT tables are created. Therefore for the file `foo-000000.ts` another `foo-000000.psi` is created. And this file constains the "missing" packets. So with a command like `cat foo-000000.psi foo-000000.ts > foo-000000.ts` the _standard_ chunk is obtained. This alternative mode could be used in case the standard HLS segments are necessary.

#### Alternatives:
The only alternative that I think it would be done is to post-process all segments to remove the PSI tables. However, this is error-prone because you don't know the number of TS packets occupied by these tables. Futhermore, if the original position of the tables are coincident with the starting of the segment, then you can't remove them. So the only solution is that the output plugin (segmenter) will do this task.

#### Additional information:
Because the work necessary to complete this action is small I think this functionality will be interesting. Then we can use HLS as a "chunk base fileformat". And we can use then tools for analyze and process only some parts of the full stream very easy. In fact all resulting segments are valid TS streams, because they're only splices of the original stream. So a very flexible and powerful technique with a minimal cost.

I hope then you (@lelegard) want to implement this. You agree?
Thank you.

lelegard:
This is no longer HLS, this is just TS slicing, cutting a raw TS file in chunks of multiples of 188 bytes. There are tools for that, not specifically for TS files.

[...]

PR #1127, opened 2022-11-24
HLS output: new "--slice-only" option

lars18th:
This implements in the "HLS output" a new parameter called "--slice-only".
Related to #1125


lars18th:
Hi @lelegard ,

The test suite in the CI fails _only_ when checking the output of the help. I've tested the output with and without the new parameter. Without the output is identical to the previous one. And with it, the output is identical when joining the segments externally (with a cat) or when reading the playlist with the HLS input.

So all tests pass!

lelegard:
I will merge this one. And implement #1128.

> The test suite in the CI fails only when checking the output of the help.

[Read this](https://tsduck.io/doxy/contributing.html#autotoc_md2) if you want to avoid CI test failure. In short, it says that when your code contribution requires a modification of the test suite, then you may fork and update the test suite accordingly and propose a pull request on both repositories. The CI on a pull request will use the test repo of the submitter if there is one. Otherwise, it uses the default test repo.

PR #1146, opened 2022-12-21
migrate from AdoptOpenJDK API to Adoptium

gdams:
The AdoptOpenJDK API has been deprecated for some time in favour of the Eclipse Temurin API


gdams:
> Thanks. The redirection from adoptopenjdk.net to adoptium.net was already automatic (see log below). But it is worth updating the script anyway.

We've been redirecting for some time now but are keen to move people away from the deprecated endpoint

PR #1151, opened 2022-12-31
Corrections in RIST plugins' comments

aplathan:
Changed a reference to a correct external library.

#### Affected components:
* tsRISTInputPlugin
* tsRISTOutputPlugin

RIST plugins have a comment that refers to an external dependancy libsrt. Corrected that to librist.

Since the changes are in code comments, there are no functional changes and thus no test were run.


lelegard:
Thanks for noticing this. An obvious copy/paste error as you may guess.

PR #1158, opened 2023-01-11
Function DtInpChannel::Read must use parameter TimeOut=-1 for blockin…

EricDeJong:
…g read mode


EricDeJong:
TsDuck expects a blocking read to the DekTec API, when no bytes are returned is is expected to be end of stream.
In the DekTec API, parameter TimeOut = -1 means block until data is available. Without TimeOut parameter, the function returns immediatly, with 0 bytes read if no data is available, causing TsDuck to exit.

lelegard:
Hi @EricDeJong 

Thanks for the code update.

The DTAPI Core Class reference manual is not quite clear about that, it is more or less implied that the version of DtInpChannel::Read() with 2 parameters (no timeout) was blocking and hangs until data is received. I no longer have any Dektec device to test but I used to have some and I do not remember having experienced immediate exit before starting the transmission on the other end. Maybe it depends on the device type or other configuration parameters.

Anyway, you are from Dektec, you must know better than I do. I will merge this.

PR #1178, opened 2023-03-03
Add new transport stream utility tslatencymonitor

amoscheungmc:
Added a new transport stream utility called 'tslatencymonitor' for monitoring the latency of two inputs by comparing their PCR input time difference.

The ‘tslatencymonitor’ utility is composed of four components:

tslatencymonitor: 
Entry point of the utility. It starts an asynchronous log, creates a core class for input processing, and sets up the utility's command line options.

tsLatencyMonitorArgs: 
Defines the arguments for the ‘tslatencymonitor’ utility. It allows the user to specify the output file name, buffer time (in seconds) to indicate the time frame of the PCR timing data list, output interval (in seconds), and loads all plugin descriptions for input plugins.

tsLatencyMonitorCore: 
The core class for input processing in 'tslatencymonitor'. It’s responsible for handling packets received from each input thread and checking if there are any PCR in the incoming packets.
If a PCR is found, the class groups it with the corresponding timestamp into a "TimingData" object and adds it to the timing data list. This list is limited by the buffer time specified by the user (default is 1s), and if the time difference between the oldest and newest data in the list exceeds the buffer time, the oldest data is removed.
Additionally, this class compares the timestamps between the two inputs to determine the latency between them. This is done by finding the difference in timestamp of the most recent PCR found in both inputs. 
After the comparison process finishes, the class will output the latency and the maximum latency.

tstslatencymonitorInputExecutor: 
This class represents an input thread that is responsible for reading packets from a single input stream and passing the packet to the core class for processing.


lelegard:
Hello @amoscheungmc,

Thank your for this interesting contribution.

I am a bit busy at the moment and I will review it in a few days.

In the meantime, could you please check the following:

- The non-regression tests raised minor issues with clang.
- Provide your real first and last name in the copyright of the source files you create, instead of mine. This is _your_ code, not mine.
- Similarly, add your name in `CONTRIBUTORS.txt`.
- Please follow the [coding guidelines](https://tsduck.io/download/docs/tsduck-coding.pdf) of the project. The whole stuff is not necessarily worth reading but I would like to emphasize the following:
  - Do not use `using namespace`, always qualify everything. This is more clear when reviewing the code. See other source files if necessary.
  - Use `}` and `else {` on two lines instead of `} else {`.
  - Use C++ casts, not old C-style casts, e.g. `double(pcrDelta)` instead or `(double)pcrDelta`.
- There is a typo in the class name `LantencyMonitorArgs`. The art of copy/paste (or auto completion) as a typo propagation engine :)
- Be sure to resynchronize your repo with `tsduck/tsduck` before updating. I am performing a few updates in TSDuck with the corresponding updates in the test suite. If you do not resynchronize the main repo first, the non-regression test on your next update will likely fail since the test suite will not match.

Thanks

lelegard:
Hi @amoscheungmc 

Your contribution has been merged into TSDuck. The user's guide has been updated with the relevant documentation (mainly a reformatted copy/paste of `tslatencymonitor --help`).

It would be nice to provide a usage example which would fit in chapter 5.1 of the user's guide. Ideally, a real-world example with the full command line and input.

Thanks again for your contribution.

amoscheungmc:
Hi @lelegard,

I really appreciate the opportunity to contribute to TSDuck.

Here is the usage example:

The `tslatencymonitor` utility is mainly used to test the real-time latency of the two streams from the same source.

In this example, two IP inputs are used, and both inputs are directly received by input plugin in `tslatencymonitor`. The buffer time is set to 2, which means the buffer time of the timing data list is 2 seconds, and the latency within 2 seconds will be trackable. The output interval is set to 3, which means there will be an output every 3 seconds. The complete command is the following:
$ tslatencymonitor --buffer-time 2 --output-interval 3 -I ip 224.1.1.1:10001 -I ip 224.1.1.1:10002
PCR1,PCR2,Latency (ms),Max Latency (ms) 
871993787,871993787,0.003,0.003 
952831712,952831712,0.096,0.096 
1034271757,1034271757,0.033,0.096 
1115207701,1115207701,0.075,0.096 
1195779574,1195779574,0.016,0.096 
1277051585,1277051585,0.02,0.096 
1358323596,1358323596,0.01,0.096


The `tslatencymonitor` utility also supports CSV output:
$ tslatencymonitor -o latency.csv -I ip 224.1.1.1:10001 -I ip 224.1.1.1:10002

PR #1192, opened 2023-03-28
Add SRT output sample based on sample-tsp.py

Marware:
#### Affected components:
Python samples

Python sample to demonstrate adding srt listener as output, which can be applied to most tsp outputs

1. All the samples `tsp.output` are `['drop']` which was confusing when I first began using the library, which will help new users
2. Use string to set input and ouput and use `split()` with `tsp.input` and `tsp.output` for better readability 

Thanks for the great job in tsduck!

### DIFF
--- a/sample/sample-python/sample-tsp-srt-out.py
+++ b/sample/sample-python/sample-tsp-srt-out.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+#----------------------------------------------------------------------------
+#
+# TSDuck sample Python application running a chain of plugins.
+#
+#----------------------------------------------------------------------------
+
+import tsduck
+
+# Create an asynchronous report to log multi-threaded messages.
+# Initial level is verbose, using time-stamped messages.
+rep = tsduck.AsyncReport(severity = tsduck.Report.Debug, timed_log = True)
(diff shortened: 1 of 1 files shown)


lelegard:
Thanks @Marware 

It is always good to have more examples. Examples help users.

I have two remarks.

1) Since this is an external contribution, could you please add a copyright with your name in the header? Just like this. It is important that authors of new files are correctly referenced.
~~~
#  TSDuck sample Python application ...
#  Copyright (c) 2023, you-first-name your-last-name
#  All rights reserved.
~~~

2) The following one is not a good idea, sorry:
> Use string to set input and ouput and use split() with tsp.input and tsp.output for better readability

Please use lists of strings. In the general case, using Python's `split()` is not the correct way of parsing a string of parameters. You have to deal with potential spaces in parameters (in service names for instance) and define a way to escape them and other exceptions. Unless you want to rewrite a shell, take advantage of being able to provide individual parameters in lists of strings, without parsing issues.

I know that there is no parsing issues in your specific sample code. However, examples must be exemplary. They are used as base code by users for other applications. You can be sure that, some day, some user will run into some spacing issue based on an example with `split()` and will complain here that it does not work. And that will take time to understand and respond.

These two remarks equally apply to your second example in #1193.

PR #1249, opened 2023-07-03
tslsdvb: fix listing tuner information for tuners with mutiple frontends

rodrigovb96:
Currently if a user tries to list all tuners with ```tslsdvb``` command in a environment where a tuner has more then one frontend and only one demux the command with fail reporting that the device is busy, that's due to the code trying to open the fd for the device mutiple times. A simple solution here is to use the abort() method to clear the file descriptors between open calls.

### DIFF
--- a/src/libtsduck/dtv/linux/tsTunerDevice.cpp
+++ b/src/libtsduck/dtv/linux/tsTunerDevice.cpp
@@ -181,6 +181,9 @@ bool ts::TunerBase::GetAllTuners(DuckContext& duck, TunerPtrVector& tuners)
             ok = false;
             tuners[index].clear();
             tuners.resize(index);
+        } else {
+            // Close all file descriptors, this it's possible to list multiple frontends in the same tuner/device.
+            tuners[index]->abort(true);
         }
     }
 
@@ -494,7 +497,7 @@ bool ts::TunerDevice::close(bool silent)
 
 void ts::TunerDevice::abort(bool silent)
 {
-    // Hord close of all file descriptors, hoping that pending I/O's will be canceled.
+    // Hard close of all file descriptors, hoping that pending I/O's will be canceled.
     // In the case of a current read operation on the dvr, it has been noticed that
     // closing the file descriptor make the read operation hang forever. We try to
     // mitigate this risk with a volatile boolean which is set around read() but


lelegard:
Hi @rodrigovb96 

There is something wrong with this PR. It marks [26 files changed](https://github.com/tsduck/tsduck/pull/1249/files). Most of these changes are old changes. It is difficult to see what is really in this PR, meaning the changes you propose. I don't understand how a PR can end up like this.

Could you please review it and resubmit it with actual differences only?
It cannot be integrated like this.
Thanks.

rodrigovb96:
Hi @lelegard , sorry for this issue here, I guess the problem is the target branch, it's pointing to develop when it should be pointed to master ( my bad ). Can you change the target branch? If it's not possible, I will close this PR and create another one with the correct target branch.

rodrigovb96:
Hi @lelegard  sorry for the confusion, the target branch was wrong, now it's correct

lelegard:
Could you please explain the exact nature of the problem?

> Currently if a user tries to list all tuners with tslsdvb command in a environment where a tuner has more then one frontend and only one demux the command with fail reporting that the device is busy, that's due to the code trying to open the fd for the device mutiple times.

Which device is opened multiple times? There should be one device per frontend. And using `tslsdvb`, the demux device is not opened.

Does `tslsdvb -v` still work with your modification? Just quickly reading the code, it shouldn't. Aborting the tuner object leaves it in an unstable state. It is still marked as open but all file descriptors are closed. When trying the read the tuner state with verbose output, it should fail.

Could you please post the output of `ls -lR dev/dvb*`?

And also the output of `tslsdvb` without your modifications?

rodrigovb96:
Hi I explained my self wrongly, but the thing here is that it's possible to have multiple frontends in a single device:
```
rodrigo@desktop:~$ ls -lR /dev/dvb*
/dev/dvb:
total 0
drwxr-xr-x 2 root root 140 Jun 30 19:30 adapter0

/dev/dvb/adapter0:
total 0
crw-rw---- 1 root video 212, 0 Jun 30 19:30 demux0
crw-rw---- 1 root video 212, 1 Jun 30 19:30 dvr0
crw-rw---- 1 root video 212, 3 Jun 30 19:30 frontend0
crw-rw---- 1 root video 212, 4 Jun 30 19:30 frontend1
crw-rw---- 1 root video 212, 2 Jun 30 19:30 net0
```
Without any change: 
```
rodrigo@desktop:~$ sudo tslsdvb 
tslsdvb: error opening /dev/dvb/adapter0/frontend1: Device or resource busy
```
using the binary with the change:
```
rodrigo@desktop:~$ sudo ./tslsdvb-mod 
0: /dev/dvb/adapter0 ("Realtek RTL2832 (DVB-T)", DVB-T)
1: /dev/dvb/adapter0:1:1:1 ("Sony CXD2837ER DVB-T/T2/C demodulator", DVB-T, DVB-T2, DVB-C)
```
with verbose:

```
rodrigo@desktop:~$ sudo ./tslsdvb-mod -v

0: /dev/dvb/adapter0 ("Realtek RTL2832 (DVB-T)", DVB-T)
   Device: /dev/dvb/adapter0/frontend0

     Signal: locked: no

     Delivery system ............................ DVB-T 
     Frequencies:
       Min ................................ 174,000,000 Hz
       Max ................................ 862,000,000 Hz
       Step ................................... 166,667 Hz
       Tolerance .................................... 0 Hz

     Capabilities: inversion auto, FEC 1/2, FEC 2/3, FEC 3/4, FEC 5/6, FEC 7/8,
       FEC auto, QPSK, 16-QAM, 64-QAM, QAM auto, transmission mode auto,
       guard interval auto, hierarchy auto, recover, mute TS

1: /dev/dvb/adapter0:1:1:1 ("Sony CXD2837ER DVB-T/T2/C demodulator", DVB-T, DVB-T2, DVB-C)
   Device: /dev/dvb/adapter0/frontend1

     Signal: locked: no

     Delivery system ............................ DVB-T 
     Frequencies:
       Min ................................. 42,000,000 Hz
       Max .............................. 1,002,000,000 Hz
       Step ......................................... 0 Hz
       Tolerance .................................... 0 Hz

     Capabilities: inversion auto, FEC 1/2, FEC 2/3, FEC 3/4, FEC 5/6, FEC 7/8,
       FEC auto, QPSK, 16-QAM, 32-QAM, 64-QAM, 128-QAM, 256-QAM, QAM auto,
       transmission mode auto, guard interval auto, hierarchy auto,
       2nd generation, mute TS
```

and also having another look  I thin that the name of the device in for the second frontend is not right ```/dev/dvb/adapter0:1:1:1```

lelegard:
Hi @rodrigovb96,

I do not have any tuner like yours, one demux, multiple frontends. I did some debug and trace on a standard tuner, one frontend, one demux.

I confirm my initial objections. I think the analysis is not correct and the solution is not correct either.

Concerning the issue, during `tslsdvb`, the tuners are open in "info only" mode. This means that, internally in the Linux implementation, the frontend is open but neither the demux nor the dvr. So, the single demux is not open twice, it is not open at all.

It seems that the driver of your tuner raises an error when you simply open frontend1 while frontend0 is already open. All tuners usually allow multiple open operations. They do not allow multiple tune or modification. But you can still open and read the characteristics of the device. This is why there is an "info only" mode in the tuner open operation.

In practice, it is quite common to tune and receive in one process (`tsp -I dvb`) and `tslsdvb -v` in another process to see the actual tuning parameters. With DVB-T, you can specify random tuning parameters and only the frequency matters. Then, when the tuner is cooperative (and this is not always the case), it reports the actual tuning parameters, that you can see through `tslsdvb -v`.

And this is precisely where the solution is wrong. When you abort the tuner in the initial list phase, it closes the frontend device (as already mentioned, the demux and dvr devices were not open). Then, in the second phase, reading the actual tuning parameters fails because the frontend was closed.

There is no error message on purpose because there are so many inconsistencies between tuner drivers, some can report such or such information while others fail, that the choice of `tslsdvb -v` is to display whatever can be displayed from the current tuner driver and don't complain about the rest. Otherwise, we would be flooded by error messages from non-cooperative drivers.

The problem with your solution is that `tslsdvb -v` will no longer display the actual tuning parameters, never, with no tuner driver.

Therefore, I cannot accept that PR.

Issue #1229, opened 2023-06-10
[BUG] Debian package declares incorrect libsrt version dependency


ajkq:
#### Bug description:
The TSDuck deb package compiled on Debian 11 and 12 declares a dependency on either libsrt1.4-openssl or libsrt1.5-openssl.
However, on Debian 11 only libsrt1.4-openssl is available.
On Debian 12, only libsrt1.5-openssl is available.

This causes TSDuck 3.34-3197.debian11 that was originally installed on Debian 11 to continue to function after an upgrade to Debian 12, since the library remains installed.
Once the TSDuck package is upgraded to 3.34-3197.debian12 with dependencies satisfied by executing `apt install --fix-broken`, tsp fails to execute with the below error message. Since the package declares a dependency on either libsrt1.4-openssl or libsrt1.5-openssl, the existing libsrt1.4-openssl package from before the upgrade causes dpkg to not complain about the missing package the binary is linked against.
`tsp: error while loading shared libraries: libsrt.so.1.5: cannot open shared object file: No such file or directory`

#### How to reproduce:
1. Install Debian 11 with 'Standard System utilities' and 'SSH server' selections
2. ```
   apt install curl
   curl -LJO https://github.com/tsduck/tsduck/releases/download/v3.34-3197/tsduck_3.34-3197.debian11_amd64.deb
   dpkg -i tsduck_3.34-3197.debian11_amd64.deb
   ```
   dpkg shows dependency problems
3. `apt install --fix-broken` installs libpcsclite1, libsrt1.4-openssl
4. Observe `tsp --version` working correctly
5. Upgrade to Debian 12
6. `dpkg -i tsduck_3.34-3197.debian12_amd64.deb` (See below how package was built)
7. dpkg shows a dependency error on librist
8. `apt install --fix-broken` installs libcjson1, libmbedcrypto7, librist4, but not libsrt1.5-openssl
9. Observe `tsp --version` failing to execute due to missing libsrt.so.1.5

#### Expected behavior:
Debian 11 package should only declare a dependency on libsrt1.4-openssl, Debian 12 on libsrt1.5-openssl. This would cause dpkg to show an error message about unmet dependencies, and the command `apt install --fix-broken` could be used to automatically install missing dependencies.

Packages compiled on either OS version list the same dependencies.
```
dpkg -I tsduck_3.34-3197.debian11_amd64.deb  | grep libsrt
[...]

PR #1252, opened 2023-07-05
Fix typos in Makefile leading to (syntax) error

AlfonZ42:
#### Related issue (if any):
Not really. Although I believe the typos being fixed here were introduced while fixing #1229.

#### Affected components:

Makefile

My editor showed somewhat wonky syntax highlighting of _Makefile_. That lead to finding missing `'`s in targets _deb-tools-control_ and _deb-dev-control_.

Running those targets directly (with some local workarounds due to not-quite-appropriate environment for them) resulted in
```
bash: -c: line 8: unexpected EOF while looking for matching `''
```

I believe running `make NOPCSC=1 deb` would show an error in an appropriate environment.


lelegard:
>I believe running make NOPCSC=1 deb would show an error in an appropriate environment.

Definitely. Just another silly mistake from me. I will merge this one.

Issue #1250, opened 2023-07-03
[BUG] Cross compilation tries to execute target bin


ankitp86:
#### Bug description:
Cross compile version v3.34-3197 fails due to build process trying to use target compiled bin on host PC; first cause of issue is tsxml. The host (where I'm doing the compilation) is x86, the target (where tsduck is intended to run) is arm7.

It seems that the build process first builds tools, then executes these tools to generate xml files. This would fine where host and target are the same arch. In a cross compilation environment, the tools intended to be executed on the host must be compiled using host sdk.

Note that I have been able to cross compile tsduck in similar way on older version (3.18-1236) successfully.

#### How to reproduce:
```
export SDK_ROOT=/home/ankit/sdks/sdk-1.2.1_gcc_6.x
export SDK_PREFIX=arm-unknown-linux-gnueabihf

make -j8                                        \
    CROSS_PREFIX=${SDK_ROOT}                    \
    CROSS_TARGET=${SDK_PREFIX}                  \
    TARGET_ARCH=armv7l MAIN_ARCH=arm            \
    NOTEST=1 NOCURL=1 NORIST=1 NOPCSC=1 NOSRT=1 \
    NOHIDES=1 NOVATEK=1 NODEKTEC=1 NOEDITLINE=1 \
    cross
```

#### Expected behavior:
The build process to succeed with Target binaries/library.

#### Errors and logs:
```
Activating Git LFS at user level
Activating Git LFS at repository level
  [DEP] base/xml/tsxmlUnknown.cpp
  [DEP] base/xml/tsxmlTweaks.cpp
  [DEP] base/xml/tsxmlText.cpp
...
  [GEN] tsduck.h
  [CXX] dtv/codec/tsAC3Attributes.cpp
  [CXX] crypto/tsAES.cpp
  [CXX] dtv/signalization/tsATSCMultipleString.cpp
...
  [LD] /home/ankit/repos/tsduck/bin/release-arm-NB-ANKIT/objs-libtsduck/alltables.o
  [LD] /home/ankit/repos/tsduck/bin/release-arm-NB-ANKIT/objs-libtsduck/alldescriptors.o
  [LD] /home/ankit/repos/tsduck/bin/release-arm-NB-ANKIT/objs-libtsduck/allcharsets.o
  [LD] /home/ankit/repos/tsduck/bin/release-arm-NB-ANKIT/objs-libtsduck/allplugins.o
[...]

PR #1253, opened 2023-07-06
add makefile variable NATIVEBINDIR to be able to execute native tsxml…

MMachmerth:
I have read the bugreport #1250 and I’m although using cross compilation feature of TSDuck. Therefore, I would prefer option 2: “Force the usage of a natively installed tsxml”. The cross compilation is integrated in a Yocto-System which handles the build of the native tsxml to and I needed on option to be flexible to set the directory where the native tsxml is located.
This pull request adds a makefile variable NATIVEBINDIR which defaults to BINDIR if is not set. The tsxml used in the build process is executed in this location.


lelegard:
Not merged as such, kept the idea of NATIVEBINDIR but in a slightly different way.
See comments in #1250

PR #1306, opened 2023-10-03
update libtsduck pkg-config to respect makevar CXXFLAGS_STANDARD

### LINKED ISSUE #1305: [BUG] Pkg-Config emits incorrect cflags
#### Bug description:
When compiled and installed using 
```
make CXXFLAGS_STANDARD=-std=c++20 && sudo make CXXFLAGS_STANDARD=-std=c++20 install
```

pkg-config emits the following flags: `--std=c++11 -I/usr/include/tsduck`. The flag `--std=c++11` should not be emitted.

#### How to reproduce:
1. Clone the repo
2. `make CXXFLAGS_STANDARD=-std=c++20`
3. `sudo make CXXFLAGS_STANDARD=-std=c++20 install`
4. `pkg-config --cflags tsduck`

#### Expected behavior:
`tsduck` could choose to emit the updated std flags, or just not to emit standard flags entirely. I understand the latter may be impossible given #1050.

#### Environment:
 - OS: Linux
 - OS version: Ubuntu 22.04
 - TSDuck full version: 3.36-3297
 - Installation type: Compiled from source, `make install`ed

#### Additional information:
The current pkg-config file confuses the meson build system, and results in projects that include tsduck to be unable to use a c++ standard other than 11.

fennewald:
#### Related issue (if any):
Closes #1305 

#### Affected components:
libtsduck pkg-config will now output `--std=c++14` and `--std=c++20` when compiled with `CXXFLAG_STANDARD=-std=c++14` and `CXXFLAG_STANDARD=-std=c++20`, respectively.

* Updates the package-config file to use a placeholder instead of a literal flag
* Updates the libtsduck makefile to replace the placeholder with the flag as set by make

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@
 #  - NOEDITLINE : No interactive line editing, remove dependency to libedit.
 #  - NOGITHUB   : No version check, no download, no upgrade from GitHub.
 #  - NOHWACCEL  : Disable hardware acceleration such as crypto instructions.
+#  - NOPCSTD    : Remove the std=c++11 flag from libtsduck's pkg-config file.
 #
 #  Options to define the representation of bitrates:
 #
--- a/src/libtsduck/config/Makefile
+++ b/src/libtsduck/config/Makefile
@@ -100,6 +100,7 @@ install-devel:
 	    $(if $(NOPCSC),-e 's|libpcsclite||g') \
 	    $(if $(NOCURL),-e 's|libcurl||g') \
 	    $(if $(NOEDITLINE),-e 's|libedit||g') \
+	    $(if $(NOPCSTD),-e 's| *--std=[^ ]* *| |g') \
 	    $(if $(NOVATEK),-e 's|libusb-1.0||g') \
 	    $(if $(OPENBSD)$(NETBSD),-e 's| *-ldl *| |g') \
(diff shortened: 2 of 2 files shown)


fennewald:
Note: I have not run the test suite, as this change does not touch any code

lelegard:
Hi @fennewald 

Not acceptable for me, as explained in my response in #1305. When you install a software, you install it for everyone, not only you. Your application probably needs C++20 but another one many need C++14 or C++11 or whatever. 

Currently, I don't see any other acceptable solution. Let's discuss the options in #1305.

fennewald:
Hi @lelegard! I've given another attempt using makevars, and explained some of my reasoning in the linked issue. If this is not to your liking, then I think the only option is that I continue to maintain my fork.

Regardless of your decision, thank you for your response and work maintaining tsduck!

lelegard:
Please use something like this, in case the actual standard is upgraded one day:
~~~
  $(if $(NOPCSTD),-e 's| *--std=[^ ]* *| |g') \
~~~

lelegard:
I merged it. To be clear with applications which use that type of installation but forget to specify some `-std` option, the root header now aborts the compilation if C++11 is not used, at least.

Additionally, I normalized old inconsistent usages of `-std` and `--std` options. Both seem to be accepted by gcc and clang (otherwise, it would have failed a long time ago) but `-std` is the officially documented one.

PR #1312, opened 2023-10-10
EITGenerator: enforce DVB 25 ms gap between consecutive sections

cmassiot:
#### Related issue (if any):
DVB states that a 25 ms gap should be enforced between sections of the same PID, table ID and table ID extension.
#### Affected components:
EITGenerator

This patch fixes plug-in eitinject.

### DIFF
--- a/src/libtsduck/dtv/epg/tsEITGenerator.cpp
+++ b/src/libtsduck/dtv/epg/tsEITGenerator.cpp
@@ -13,6 +13,9 @@
 #include "tsBCD.h"
 
 
+/* DVB specifies at least 25 ms. */
+#define SECTION_GAP 30
+
 //----------------------------------------------------------------------------
 // Constructor.
 //----------------------------------------------------------------------------
@@ -36,6 +39,9 @@ ts::EITGenerator::EITGenerator(DuckContext& duck, PID pid, EITOptions options, c
     _packetizer(_duck, _eit_pid, this),
     _services(),
     _injects(),
+    _last_tid(0),
+    _last_tidext(0),
+    _last_index(0),
     _obsolete_count(0),
     _versions()
 {
@@ -75,6 +81,7 @@ void ts::EITGenerator::reset()
     for (size_t i = 0; i < _injects.size(); ++i) {
         _injects[i].clear();
     }
+    _last_tid = 0;
     _obsolete_count = 0;
     _versions.clear();
 }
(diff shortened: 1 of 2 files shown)


cmassiot:
Hi Thierry, and thanks for your work. DVB states that there should be a gap between the last byte of a section and the first byte of the next section with the same {PID, TID, TIDEXT}. The problem is that you have no way to know exactly when a section will be inserted, you only know the earliest date, and you certainly don't know the date of the last byte if the section is big and split into several TS packets (which can be the case for EIT).

It shouldn't be a problem if you have a high number of padding packets, but if the bandwidth is scarce, a section could be delayed signficantly. Besides you have no control over it. That is why I chose the current method that is 100% certain to enforce the section gap, but I leave it to you.

lelegard:
OK, I understand your concern about the gap _after the last packet_.

>  That is why I chose the current method that is 100% certain to enforce the section gap, but I leave it to you.

The timing issue of first vs. last packet of a section is the same with the two methods. The `provideSection()` method is called when the packetizer needs a section. You provide the section and then the packetizer will spead it over several TS packets in the next future. Therefore, when you reschedule sections in `provideSection()`, the rescheduled time is based on the time of the first packet of the current section, not the last one.

lelegard:
I think that the packetizer is the only object which is aware of the exact moment where the last packet of a section is injected.

Maybe I could enhance `SectionProviderInterface` to add an optional virtual method through which the packetizer will notify the section provider of the last packet of each section. In this method, you would ideally reschedule the sections of the same tid/tidext.

However, a corner case remains (also present in your PR). Let's assume that there is no enqueued section of the same tid/tidext and, consequently, nothing to reschedule. However, within the next 25 ms, a new section of the same tid/tidext is enqueued, and the queues are not much populated. Then, the section will likely be injected less than 25 ms after the previous one.

A general solution would be to keep a map of last injection time of last packet per tid/tidext. However, this would increase the burden even more. The EIT generator is very demanding in terms of resources, despite a few optimizations I implemented. An efficient and fully DVB compliant implementation would maybe require a complete rewrite. But I have no time for that.

cmassiot:
If you look closely you'll notice that I only reschedule sections when the packetizer asks for a *new* section. That way I am sure that the *previous* section has been completely transmitted, so we have at least SECTION_GAP until the next section with identical TID,TIDEXT. In reality we have more than SECTION_GAP because the last section may have finished being transmitted earlier, but that's not a big deal in my opinion.

Of course having a callback in the packetizer to notify when a section has been fully transmitted would work to, and allow for a smaller gap.

lelegard:
> If you look closely you'll notice that I only reschedule sections when the packetizer asks for a new section. 


Sorry, my bad. I read it again more closely and realized that the rescheduling of sections concerned the _previous_ section, the completed one, not the section which is returned by the `provideSection()` callback.

lelegard:
Hi Christophe,

I merged it in the master branch. Since I rebased before merging, the PR is not automatically detected as merged in GitHub and I will have to close the PR. If you look at the master branch, you will see your commit. I never understood this behavior and don't know how to fix it. Any idea?

There is another commit on top of it with a bunch of refactoring (I am in a background process of simplifying code since I dropped any pre-C++11 support).

I also modified one thing in your code: calling `enqueueInjectSection()` while working through the list was not appropriate since the call modifies that list, including the risk of invalidating the current iterator.

lelegard:
Merged

PR #1315, opened 2023-10-13
EIT: change options to independantly select the 4 kinds

cmassiot:
#### Related issue (if any):
The current options for EITGenerator only allow to select either actual or other, and either pf or schedule. You have no way for instance to select actual pf, other pf and actual shedule but not other schedule. However this is what you are supposed to do on eg. the French DTT.

#### Affected components:
EITGenerator, eitinject tsp plug-in and tseit application.

This commit makes 4 distinct options, one for each kind. It does however change the CLI of tsp and tseit. I am unsure what the policy of the project is regarding such changes, so apply with caution.

### DIFF
--- a/src/libtsduck/dtv/epg/tsEITGenerator.cpp
+++ b/src/libtsduck/dtv/epg/tsEITGenerator.cpp
@@ -428,7 +428,7 @@ void ts::EITGenerator::setTransportStreamId(uint16_t new_ts_id)
         if (new_other || new_actual) {
 
             // Process EIT p/f.
-            if (bool(_options & EITOptions::GEN_PF)) {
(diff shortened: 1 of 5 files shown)


AlfonZ42:
Is it possible to retain old options and have them work as aliases/synonyms for newly added more specific options, e.g. similar to how default standard selection works (see chapters 2.4.2 and 2.5.2 of TSduck documentation)?

Testing the new options within the test suite would be good. However, I'm not sure whether it's better to adjust tests _98_, _104_, and  _123_, or add a new test.

lelegard:
I merged the PR and restored the previous options to avoid breaking existing scripts. Older options now behave as combinations of new options, e.g. `--actual` is restored and equivalent to `--actual-pf --actual-schedule.`

The test suite has not been modified.

PR #1327, opened 2023-10-22
Correcting properties for MPEG codecs

paulhiggs:
Previous code used different 'versions' for the same values, somewhat due to inconsistencies in ISO/IEC 14496-15 which are not corrected in development drafts


lelegard:
Thank you Paul. Accurate as usual!
I will merge this one.

PR #1372, opened 2023-11-29
Add support for custom HTTP headers in HTTP/HLS communications

willdelstrother:
#### Affected components:
HLS / HTTP input

Allow the user to set custom header/s in HTTP requests - I added this because a particular HLS stream I needed to ingest was using a custom token-based authentication scheme.

Example usage:

```
tsp -I hls --headers 'x-affiliate-auth-key: {token}' https://{url}/index.m3u8 -O file  /tmp/test.mp4
```

Several headers can be added with multiple `--headers` entries.

### DIFF
--- a/src/libtsduck/base/network/tsWebRequest.cpp
+++ b/src/libtsduck/base/network/tsWebRequest.cpp
@@ -195,6 +195,9 @@ void ts::WebRequest::setArgs(const ts::WebRequestArgs& args)
     if (args.useCompression) {
         enableCompression();
     }
+    for (const auto& it : args.headers) {
+        setRequestHeader(it.first, it.second);
+    }
 }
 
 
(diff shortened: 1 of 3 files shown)


lelegard:
Thanks @willdelstrother, good idea.

May I add your name in the list of contributors? Which name should I use ([a full real name please](https://tsduck.io/doxy/contributing.html))?

willdelstrother:
Sure, that's fine, although realistically I don't think I'll be able to
make contributions at all regularly.

My name is William del Strother.

Thanks

lelegard:
Don't worry about the check failures, this is because of the new options. The test suite checks that the help is not broken and verifies that the help did not change. This is also a good way to track new options to update the doc. I have a procedure to realign the test suite.

And forget about my last question, you already added your full name in the commit.

PR #1380, opened 2023-12-11
remove non-ACSII characters

paulhiggs:
#### Related issue (if any):
No specific issue, just a standards industry colleague informing me that tgge comment in tSPSI.h and the description in tsPSI.names contained non-ascii characters

#### Affected components:
tsPSI

Replace non-ASCII hyphen with ASCII hyphen for improved printout


lelegard:
It looks like copy/paste from a PDF file 😉

PR #1399, opened 2023-12-31
Follow-up on b7fcdbc

rubenochiavone:
Follow-up on b7fcdbc.

### DIFF
--- a/sample/install/install-tsduck.ps1
+++ b/sample/install/install-tsduck.ps1
@@ -1,6 +1,6 @@
 ﻿#-----------------------------------------------------------------------------
 #
-#  Copyright (c) 2021, Thierry Lelegard
+#  Copyright (c) 2021-2024, Thierry Lelegard
 #  BSD-2-Clause license, see LICENSE.txt file or https://tsduck.io/license
 #
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


lelegard:
Hi @rubenochiavone,

Thanks for taking the time to look into this. 

The `__copyright__` string in the Python module was indeed overlooked and I will update it independently.

The PowerShell script files in `scripts/install-wintools`, on the other hand, are exact copies of their counterparts in a personal repo with many Windows installation scripts. This is [explained here](https://github.com/tsduck/tsduck/blob/master/scripts/install-wintools/README.md). I prefer to keep them aligned with the original repo.

However, this was to opportunity to notice that their license lines were inadvertently updated by a global `sed` command that I ran months ago on all TSDuck source files. I just reverted them.

As an independent remark, I noticed that all checks failed in your PR. This is because you have your own copy of the `tsduck-test` repo and it is not up-to-date. During a PR, the CI workflow checks if the author of the PR has his own copy of the `tsduck-test` repo, in which case it is used instead of `tsduck/tsduck-test`. Thus, if the author of the PR has made updates which impact the test suite, the two can be consistently used. The constraint is that, if you have your own copy of `tsduck-test`, you must synchronize it before submitting the PR. Otherwise, the CI workflow may use an obsolete version of the test suite and the CI will fail. This is exactly what happened in your PR.

PR #1408, opened 2024-01-05
Add code coverage reports generation

svlobanov:
#### Related issue (if any):
No related issue

#### Affected components:
github actions workflows

Code coverage reports are useful for adding new tests and finding bugs. In some cases, it is easier to download a report from github than running it locally (proposed workflow uploads coverage reports as github artifact). Also, it might be useful for new pull requests to decide whether more tests required.

You can see the report and workflow execution here: https://github.com/svlobanov/tsduck/actions/runs/7426865570

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,91 @@
+#-----------------------------------------------------------------------------
+#
+#  TSDuck - The MPEG Transport Stream Toolkit
+#  Copyright (c) 2005-2024, Thierry Lelegard
+#  Copyright (c) 2024, Sergey Lobanov
+#  BSD-2-Clause license, see LICENSE.txt file or https://tsduck.io/license
+#
+#  GitHub Actions configuration file : Code Coverage
+#
+#  The workflow is triggered on push or pull request, for master branch only.
+#  Manual dispatch from the GitHub UI is also allowed.
+#
+#  TSDuck is built on Ubuntu, using C++17 and C++20 levels of standards.
+#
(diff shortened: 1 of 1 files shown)


lelegard:
Hi @svlobanov 

Thanks for this contribution. I merged it.

However, I updated the workflow to be run on manual dispatch only. Running a workflow automatically on push and PR is useful to issue a binary diagnostic: pass or fail (compilation error, test error, security issue). Code coverage is a quality diagnostic. There is no need to run it every time.

I will consider publishing the coverage reports on the Web site, alongside the doygen-generated documentation. In that case, I will run it as part of the nightly build (this is the workflow which builds daily binaries and documentation and pushes them to tsduck.io).

svlobanov:
Hi @lelegard. Thanks for accepting this PR. Just one technical thing regarding `apt` vs `pip3` for gcovr install. pip3 installs gcovr v6, apt installs gcovr v5. gcovr v5 doesn't have function calls coverage which is very useful (at least from my point of view).
Anyway, it's not a big issue due to ubuntu 24.04 will be released in several months and after some time microsoft will use it as ubuntu-latest on github.

lelegard:
> pip3 installs gcovr v6, apt installs gcovr v5

On my Ubuntu 23.10, apt installs gcovr 6.0. Therefore, I did not notice the difference which applies on GitHub runners with Ubuntu 22.04.

Anyway, for the sake of consistency, it is better to use the native package manager of the operating system.

Running the command on my local machine (with gcovr 6.0), I was not convinced by the function calls coverage. Either I misinterpret the results or there is something wrong in the analysis.

PR #1415, opened 2024-01-11
Fixed use-of-uninitialized-value related to init of struct stat

svlobanov:
This potential issue was found by MemorySanitizer (msan). Whether the actual problem occurs depends on the compiler, OS, and other factors.

msan output related to this fix:

~~~
$ ./tsp
==432620==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7fcfde24e8ac in ts::TSFile::seekCheck(ts::Report&) libtsduck/dtv/transport/tsTSFile.cpp:411:18
    #1 0x7fcfde24a9c6 in ts::TSFile::openInternal(bool, ts::Report&) libtsduck/dtv/transport/tsTSFile.cpp:366:10
    #2 0x7fcfde24d542 in ts::TSFile::openRead(std::__1::__fs::filesystem::path const&, unsigned long, unsigned long, ts::Report&, ts::TSPacketFormat) libtsduck/dtv/transport/tsTSFile.cpp:151:12
    #3 0x7fcfde25980d in ts::TSFileInputArgs::openFile(unsigned long, unsigned long, ts::Report&) libtsduck/dtv/transport/tsTSFileInputArgs.cpp:149:31
    #4 0x7fcfde259fbd in ts::TSFileInputArgs::open(ts::Report&) libtsduck/dtv/transport/tsTSFileInputArgs.cpp:187:14
    #5 0x7fcfdd91fc44 in ts::FileInputPlugin::start() libtsduck/plugins/plugins/tsFileInputPlugin.cpp:37:18
    #6 0x7fcfde2b3bd2 in ts::TSProcessor::start(ts::TSProcessorArgs const&) libtsduck/plugins/apps/tsTSProcessor.cpp:179:30
    #7 0x555c0ab44103 in MainCode(int, char**) tstools/tsp.cpp:147:17
    #8 0x7fcfddc45a52 in MainWrapper(int (*)(int, char**), int, char**) libtsduck/base/app/tsMain.cpp:35:16
    #9 0x555c0ab4282d in main tstools/tsp.cpp:22:1
    ...

  Uninitialized value was stored to memory at
    #0 0x7fcfde24c6cb in ts::TSFile::openInternal(bool, ts::Report&) libtsduck/dtv/transport/tsTSFile.cpp:363:14
    #1 0x7fcfde24d542 in ts::TSFile::openRead(std::__1::__fs::filesystem::path const&, unsigned long, unsigned long, ts::Report&, ts::TSPacketFormat) libtsduck/dtv/transport/tsTSFile.cpp:151:12
    #2 0x7fcfde25980d in ts::TSFileInputArgs::openFile(unsigned long, unsigned long, ts::Report&) libtsduck/dtv/transport/tsTSFileInputArgs.cpp:149:31
    #3 0x7fcfde259fbd in ts::TSFileInputArgs::open(ts::Report&) libtsduck/dtv/transport/tsTSFileInputArgs.cpp:187:14
    #4 0x7fcfdd91fc44 in ts::FileInputPlugin::start() libtsduck/plugins/plugins/tsFileInputPlugin.cpp:37:18
    #5 0x7fcfde2b3bd2 in ts::TSProcessor::start(ts::TSProcessorArgs const&) libtsduck/plugins/apps/tsTSProcessor.cpp:179:30
    #6 0x555c0ab44103 in MainCode(int, char**) tstools/tsp.cpp:147:17
    ...

  Uninitialized value was created by an allocation of 'st' in the stack frame of function '_ZN2ts6TSFile12openInternalEbRNS_6ReportE'
    #0 0x7fcfde24a0b0 in ts::TSFile::openInternal(bool, ts::Report&) libtsduck/dtv/transport/tsTSFile.cpp:196

SUMMARY: MemorySanitizer: use-of-uninitialized-value libtsduck/dtv/transport/tsTSFile.cpp:411:18 in ts::TSFile::seekCheck(ts::Report&)
~~~

#### Related issue (if any):
no related issue

#### Affected components:
TSDuck library

This patch properly inits st (stat) structure

### DIFF
--- a/src/libtsduck/base/system/tsFileUtils.cpp
+++ b/src/libtsduck/base/system/tsFileUtils.cpp
@@ -320,10 +320,10 @@ fs::path ts::TempFile(const UString& suffix)
 ts::Time ts::GetFileModificationTimeUTC(const UString& path)
 {
 #if defined(TS_WINDOWS)
-    ::WIN32_FILE_ATTRIBUTE_DATA info;
+    ::WIN32_FILE_ATTRIBUTE_DATA info {};
     return ::GetFileAttributesExW(path.wc_str(), ::GetFileExInfoStandard, &info) == 0 ? Time::Epoch : Time::Win32FileTimeToUTC(info.ftLastWriteTime);
 #else
-    struct stat st;
+    struct stat st {};
     return ::stat(path.toUTF8().c_str(), &st) < 0 ? Time::Epoch : Time::UnixTimeToUTC(st.st_mtime);
 #endif
 }
--- a/src/libtsduck/dtv/transport/tsTSFile.cpp
+++ b/src/libtsduck/dtv/transport/tsTSFile.cpp
@@ -352,7 +352,7 @@ bool ts::TSFile::openInternal(bool reopen, Report& report)
     }
 
     // Check if this is a regular file.
-    struct stat st;
+    struct stat st {};
     if (::fstat(_fd, &st) < 0) {
         report.log(_severity, u"cannot stat input file %s: %s", {getDisplayFileName(), SysErrorCodeMessage()});
         if (!_std_inout) {


lelegard:
Hi @svlobanov 

Thanks for the PR. It is always good to have all variable initialized, even when their initial value is (supposedly) never read and immediately overwritten. I consider this as good software engineering practices.

However, in that specific case, this is a false positive or, more precisely, the symptom of a bug in the libc. The `struct stat` parameter in the `stat()` and `fstat()` functions is an output parameter. It is supposed to be written by `stat()`, not read and certainly not read before being assigned by this function. Anyway, thanks, I will merge this one.

lelegard:
>  The struct stat parameter in the stat() and fstat() functions is an output parameter. It is supposed to be written by stat(), not read 

My analysis was incorrect here. I carefully looked at the source line as reported by the sanitizer:
~~~
    struct stat st;
    if (::fstat(_fd, &st) < 0) {
        // error processing...
        return false;
    }
    _regular = S_ISREG(st.st_mode);  <== "uninitialized" error here
~~~

This is not an attempt of `fstat()` to read the structure before writing to it. This is the normal use of the returned value by the application.

I think we hit here a limitation of the memory sanitizer tool. Because, if it was true that `st.st_mode` was uninitialized, then it would be a huge bug in the libc and the kernel. This kind of thing would have been discovered a long time ago.

So, my hypothesis is the following: `fstat()` is a libc function (user mode) which calls the syscall `stat()` (kernel mode). The `struct stat` is written by `stat()` from kernel mode only. It is pretty obvious that the memory sanitizer cannot track kernel code. It can only track memory access from user mode. Therefore, it is unable to determine that `st.st_mode` was indeed initialized, but outside its control.

This is an interesting example because it demonstrates that the address sanitizer cannot, by design, track all memory accesses. This is not a bug which can be fixed in the tool. This is an intrinsic limitation. In that specific case, it can be hidden by pre-initializing the variable, as you did in the PR (and this is good engineering practice anyway). However, there may be other corner cases where the address sanitizer cannot detect valid memory accesses.

Therefore, I will remove the asan test from the continuous integration workflow. This workflow is binary, pass of fail. A failure there shall indicate a real error. We cannot risk to reject a push or PR because of a false positive in the sanitizer. This test is more a qualitative test which needs a human interpretation. I will create a separate workflow with manual trigger, just like the code coverage workflow you previously submitted.

svlobanov:
Hi @lelegard. This particular PR was found by msan (memory sanitizer), but PR https://github.com/tsduck/tsduck/pull/1414 is about asan (address sanitizer). They are very different, use different methods to detect issues so let's separate this one and asan PR.

I will double check this issue regarding fstat(), maybe I use msan incorrectly (it requires many changes in build process to use msan).

svlobanov:
@lelegard I have checked msan and fstat() with a very simple code. msan normally works with fstat() so I will continue investigate why msan generates an use-of-uninitialized-value in tsduck code (most probably because of incorrect compilation/linker flags)

My example (fstat_test2.cpp, fstat() is called only if argc>2):
```
#include <stdio.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>
 
int main(int argc, char *argv[]) 
{ 
    struct stat st;       
    int fd = ::open(argv[1], O_RDONLY);
    if(fd<0)
        return -1;

    if(argc>2)
        ::fstat(fd, &st);

    if(S_ISREG(st.st_mode))
        printf("File is a regular file\n"); 

    else
        printf("File is a something else\n"); 
 
    ::close(fd); 
 
    return 0; 
}
```

build: `clang++ -fsanitize=memory -fsanitize-memory-track-origins -fno-omit-frame-pointer -O0 -g3 fstat_test2.cpp`

~~~
$ ./a.out a.out 1
File is a regular file

$ ./a.out a.out
==447677==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x5617d14349af in main /home/sergey/b/fstat/fstat_test2.cpp:16:8
    #1 0x7f4ec68461c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #2 0x7f4ec6846284 in __libc_start_main csu/../csu/libc-start.c:360:3
    #3 0x5617d13ac2c0 in _start (/home/sergey/b/fstat/a.out+0x222c0) (BuildId: 44c27ce1e92f8a545e37904c0b1d75dc70407261)

  Uninitialized value was created by an allocation of 'st' in the stack frame of function 'main'
    #0 0x5617d14343c0 in main /home/sergey/b/fstat/fstat_test2.cpp:7

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/sergey/b/fstat/fstat_test2.cpp:16:8 in main
Exiting
~~~

lelegard:
So your example demonstrates that msan does not work with `fstat()` because your `if(S_ISREG(st.st_mode))` is flagged as "use of uninitialized value" while it has been obviously initialized from kernel mode by the `stat()` syscall.

I do not think that this is a problem of compilation flags or a bug in msan. This is a very logical limitation of a userland tool. We just have to accept it.

This also mean that, while msan is a useful tool to track memory issues, just like asan or valgrind (my favorite), it is not fully deterministic and its success rate cannot be 100%, meaning that false positive are inevitable.

And there is no really efficient solution.

On one hand, pre-initializing all variables (as in this PR) ensures that all variables are preset in a deterministic state. This mutes the msan error.

However, on the other hand, pre-initializing all variables also partly kills the usefulness of msan. Let me explain this. In TSDuck or your code, we call a function which updates an output variable. Then we use that value. Let's assume that we have a bug and the function is not called in some cases but we still use the value. Without pre-initializing the variable (and assuming the function is not a syscall), msan will detect the uninitialized value. This is why it is useful. When we pre-initialize the variable, msan will never find the error.

So, there is no perfect solution, both have issues. However, I do prefer the pre-initialization because it has a deterministic behaviour. If it fails once, we detect and fix the bug. If it "works by accident" once, it will always "work by accident".

Concerning GitHub workflows, I consider the "continuous integration" one as a validation step. So, it must be reliable and produce a pass/fail status. No false positive is acceptable. Quality tests such as code coverage and memory analyzer, which produce a non-binary status (x% coverage) or which many produce false positive from time to time, shall not be in the CI critical path. They must be ready to be run when necessary, not all the time.

svlobanov:
> So your example demonstrates that msan does not work with fstat() because your if(S_ISREG(st.st_mode)) is flagged as "use of uninitialized value" while it has been obviously initialized from kernel mode by the stat() syscall.

It works, please check both outputs. When I run "./a.out a.out 1" argc=3, fstat() is executed and there is no msan error. 
msan error happens when I run "./a.out a.out", in this case argc=2 and fstat() is not executed. 

Also I've checked msan source code, they intercept fstat() that's why I continue investigating msan error in tsduck code

lelegard:
> When I run "./a.out a.out 1" argc=3, fstat() is executed and there is no msan error.

Agreed. Sorry, I looked too quickly over this one.

I wonder how msan can trace memory access from kernel code. `fstat()` is a userland libc function but `stat()` is a syscall.

svlobanov:
> I wonder how msan can trace memory access from kernel code.

They use interceptor for this, it looks hacky but in real life it works (google needs it because they worry of chrome security).
 
https://github.com/llvm/llvm-project/blob/f02b7770bb04b88455804b371dba1f44a86b90da/compiler-rt/lib/msan/msan_interceptors.cpp#L755-L767

svlobanov:
> Concerning GitHub workflows, I consider the "continuous integration" one as a validation step. So, it must be reliable and produce a pass/fail status. No false positive is acceptable. 

Regarding to gcc/llvm asan and llvm msan - both are bit accuracy so in ideal conditions there should be no false positive results, but in practice I can see two ways:
1. asan - issues in external libraries (asan reports in test-130 if use some versions of openssl so I had to disable it when run tests on asan build)
2. msan - calls like fstat() if llvm (google) guys not implemented interceptor. I still think that false positive report provided in this PR happens because I incorrectly use msan for tsduck.

I understand that you don't want to fight with potential false-positive sanitizers reports and implement workarounds/suppressions as a part of CI-process so I propose to use sanitizers during "fuzzing compaings" (when the fuzzing tests are implemented). Also it might be useful to run *san pipeline (even without fuzzing) from time to time (e.g. before "official" release)

As a reference, Wireshark project runs fuzz+asan (https://gitlab.com/wireshark/wireshark/-/blob/490d0514d97f0a24185a58d8bf8558290b4b5528/.gitlab-ci.yml ) by schedule: https://gitlab.com/wireshark/wireshark/-/pipeline_schedules

PR #1416, opened 2024-01-12
Fixed undefined behaviour in pcrcopy plugin that cause real issue with label processing

svlobanov:
#### Related issue (if any):
no related issue

#### Affected components:
pcrcopy plugin

If --reference-label is not set by user, then _ref_label sets to 32 (TSPacketLabelSet::MAX+1)
by ts::PCRCopyPlugin::getOptions().
ts::PCRCopyPlugin::processPacket() calls pkt_data.hasLabel(_ref_label) that calls
CompactBitSet::test() that uses 'int_t(1)<<pos' code, where int_t is uint32_t and pos=32
1<<32 is undefined by C++ standard for 32-bit unsinged integer. In real world 1<<32 usually
equals to 1 so it means that pkt_data.hasLabel(32) is equal to pkt_data.hasLabel(0)

Real issue can be caused by the following command (reproduces on macos arm64 clang-15 and
linux x86_64 gcc-12):

~~~
% tsp --verbose -I file ../tsduck-test/input/test-092.ts -P filter -p 0x140 --set-label 0 -P pcrcopy -r 0x78 -t 0x82 -O drop
* file: initial input bitrate is 24,047,795 b/s
* pcrcopy: using PID 0x0140 (320) as PCR reference
* pcrcopy: using PID 0x0140 (320) to insert copied PCR
~~~

In this example, user asks pcrcopy plugin to copy PCR values from pid 0x78 to 0x82, but tsduck copies from
0x140 to 0x140 because label 0 was set by filter plugin.

Minimal example to check the behaviour on different systems and compilers:
~~~
% cat undef.cpp
#include <iostream>
#include <sstream>

int main(int argc, char *argv[]) {
    std::cout << (uint32_t(1) << size_t(32)) << std::endl;
    std::stringstream argv1(argv[1]);
    size_t exp;
    argv1 >> exp;
    std::cout << (uint32_t(1) << exp) << std::endl;
    return 0;
}
~~~

~~~
% clang++ undef.cpp
% ./a.out 32 # on macos arm64, clang-15
1796978288
1
% ./a.out 32 # on macos arm64, second try
1842411120
1
~~~

~~~
$ g++ undef.cpp
$ ./a.out 32 # on linux x86_64, gcc-12
0
1
$ ./a.out 32 # on linux x86_64, second try
0
1
~~~

~~~
This issue was found during tsduck-test/test-093 with UBSAN enabled:
    > libtsduck/base/types/tsCompactBitSet.h:214:66: runtime error: shift exponent 32 is too large for 32-bit type 'int_t' (aka 'unsigned int')
    >     #0 0x1048eb140 in ts::PCRCopyPlugin::processPacket(ts::TSPacket&, ts::TSPacketMetadata&) tsplugin_pcrcopy.cpp:169
    >     #1 0x1078bf8f4 in ts::tsp::ProcessorExecutor::processIndividualPackets() tstspProcessorExecutor.cpp:173
    >     #2 0x1078becf0 in ts::tsp::ProcessorExecutor::main() tstspProcessorExecutor.cpp:74
    >     #3 0x10774a3d0 in ts::Thread::mainWrapper() tsThread.cpp:345
    >     #4 0x107749f98 in ts::Thread::ThreadProc(void*) tsThread.cpp:389
    >     #5 0x18b9a5030 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x7030)
    >     #6 0x18b99fe38 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1e38)
    >
    > SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior libtsduck/base/types/tsCompactBitSet.h:214:66 in
~~~

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -521,6 +521,16 @@ else
     LDFLAGS_ASAN =
 endif
 
+# Compilation flags for code sanitizing using UndefinedBehaviorSanitizer with default optimization.
+
+ifneq ($(UBSAN),)
+    CXXFLAGS_UBSAN = -fsanitize=undefined -fno-sanitize=unsigned-integer-overflow -fno-sanitize=alignment -g3
+    LDFLAGS_UBSAN = -fsanitize=undefined -fno-sanitize=unsigned-integer-overflow -fno-sanitize=alignment -g3
+else
+    CXXFLAGS_UBSAN =
+    LDFLAGS_UBSAN =
+endif
+
 # Compilation flags for code coverage using gcov.
 
 ifneq ($(GCOV),)
@@ -553,10 +563,10 @@ LDLIBS += -lstdc++ -lpthread $(if $(MACOS)$(OPENBSD),,-lrt) -lm
 # Global compilation flags.
 # Additional flags can be passed on the "make" command line using xxFLAGS_EXTRA.
 
-CXXFLAGS = $(CXXFLAGS_DEBUG) $(CXXFLAGS_M32) $(CXXFLAGS_ASAN) $(CXXFLAGS_GCOV) $(CXXFLAGS_GPROF) $(CXXFLAGS_WARNINGS) \
+CXXFLAGS = $(CXXFLAGS_DEBUG) $(CXXFLAGS_M32) $(CXXFLAGS_ASAN) $(CXXFLAGS_UBSAN) $(CXXFLAGS_GCOV) $(CXXFLAGS_GPROF) $(CXXFLAGS_WARNINGS) \
            $(CXXFLAGS_SECURITY) $(CXXFLAGS_INCLUDES) $(CXXFLAGS_TARGET) $(CXXFLAGS_FPIC) $(CXXFLAGS_STANDARD) \
            $(CXXFLAGS_CROSS) $(CXXFLAGS_PTHREAD) $(CXXFLAGS_EXTRA)
-LDFLAGS  = $(LDFLAGS_DEBUG) $(LDFLAGS_M32) $(LDFLAGS_ASAN) $(LDFLAGS_GCOV) $(LDFLAGS_GPROF) $(CXXFLAGS_TARGET) \
+LDFLAGS  = $(LDFLAGS_DEBUG) $(LDFLAGS_M32) $(LDFLAGS_ASAN) $(LDFLAGS_UBSAN) $(LDFLAGS_GCOV) $(LDFLAGS_GPROF) $(CXXFLAGS_TARGET) \
            $(LDFLAGS_CROSS) $(LDFLAGS_PTHREAD) $(LDFLAGS_EXTRA)
 ARFLAGS  = rc$(if $(MACOS),,U) $(ARFLAGS_EXTRA)
 
--- a/src/tsplugins/tsplugin_pcrcopy.cpp
+++ b/src/tsplugins/tsplugin_pcrcopy.cpp
@@ -166,13 +166,13 @@ ts::ProcessorPlugin::Status ts::PCRCopyPlugin::processPacket(TSPacket& pkt, TSPa
     const PID pid = pkt.getPID();
 
     // Process PID switching according to labels.
-    if (pkt_data.hasLabel(_ref_label) && pid != _ref_pid && pid != PID_NULL) {
+    if ((_ref_label != (TSPacketLabelSet::MAX + 1)) && pkt_data.hasLabel(_ref_label) && pid != _ref_pid && pid != PID_NULL) {
         // Switch to a new reference PID.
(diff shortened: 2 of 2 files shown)


svlobanov:
Regarding to test-131 on "Build on macos-latest with clang, C++17". Looks like, it was some CI issue, e.g. macos host was too slow.

lelegard:
Hi @svlobanov 

Thanks for reporting this. I find this amazing. After forty years of programming and writing several millions of lines of code, I was fully convinced that shifting more than the size of the type resulted in zero, at least in C/C++. I probably took this from assembly code because this is how most ISA behave. Then, I considered C as some kind of super-assembler. I fully adopted C++ after years of programming in Ada and Java. I was very well aware of the concept of "undefined behavior" in those high-level languages. Going back to C++, I probably made the mistake to consider it as a super-high-level object-oriented assembler.

I learnt something today.

I assume that the generated code first computes "shift-count modulo int-type-size" and then use this value as shift count. I will verify in the generated code. The weird result is that, depending on the amount of shift, the operation is either a shift or a rotate.

My small test:
~~~
#include <iostream>
#include <iomanip>
int main(int argc, char* argv[])
{
    uint32_t n = argc < 2 ? 1 : std::atoi(argv[1]);
    int shift  = argc < 3 ? 0 : std::atoi(argv[2]);
    std::cout << std::hex << std::setfill('0') << std::setw(8) << (n << shift) << std::endl;
}
~~~

Test below. See how the shift becomes a rotate, probably because of the modulo.
~~~
$ for s in 28 29 30 31 32 33; do ./shift 15 $s; done
f0000000
e0000000
c0000000
80000000
0000000f
0000001e
~~~

But I have a very bad feeling now. How many times did I write code based on the assumption that shifting more than the type size resulted in zero? How many latent bugs are sleeping there, waiting to blow up?

svlobanov:
I'm glad that you learned something new from this PR.

Undefined behaviour in C and C++ (be careful, the C and C++ standards define different things undefined) is a concept that allows using the fastest CPU instructions. You know that you can disasm my or your example and see why shifting becomes rotating. BTW, you can play with compiler flags to specify cpu microarchicture, the result may be different. 
The most important thing that you should think that any undefined operation MAY produce any kind garbage (like it happens in my example on macos arm64, clang-15)

> How many times did I write code based on the assumption that shifting more than the type size resulted in zero? How many latent bugs are sleeping there, waiting to blow up?

This is not just a rhetorical question, but a solvable one. There are three ways to find undefined behaviour issues:

1. unsuppress compiler warnings related to undefined behaviour and change it to 100% deterministic, e.g. https://github.com/tsduck/tsduck/blob/b8adb67a1a53072e6206b7617b40f17835b2aa11/src/libtsduck/base/types/tsIntegerUtils.h#L855-L859 (yes, negative shift is undefined behaviour in C++, but so many application developers use this operations so compiler developers *usually* don't optimize it, but some day on some specific hardware and compiler you will catch an issue)
2. check static analyzers warnings regarding undefined behaviours (if you don't suppress it). more advanced analyzers can find more potential issues
3. Use UBSAN (https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html ), it will find undefined behaviour even if data is passed from external world (argv[], .ts file, etc). Compilers (and some static analyzers) don't generate warnings if data is external.

lelegard:
> > How many times did I write code based on the assumption that shifting more than the type size resulted in zero? How many latent bugs are sleeping there, waiting to blow up?
>
> This is not just a rhetorical question, but a solvable one. There are three ways to find undefined behaviour issues:

Agreed if we can reproduce the invalid values, possibly dynamically, using external values. That is how you found that bug.

My fear is more about constructs such as "non-zero-value << shift" where "shift" is always below the size limit with valid input. If I based error processing on the assumption that the result would be zero on invalid input, we have a problem. In your PR, the input was valid and the error was in the processing of valid input. With invalid input, fuzzing may help of course. But the coverage cannot be 100% (it is infinite).

svlobanov:
> My fear is more about constructs such as "non-zero-value << shift" where "shift" is always below the size limit with valid input. 

According to C++ standards, even "-1 << 5" is undefined behaviour (you can find exact references here: https://stackoverflow.com/a/3785121/21113842 ), but in real world it works as you expected, at least it works as you expected on x86_64 and arm64 linux and macos, gcc and clang.

lelegard:
>  even "-1 << 5" is undefined behaviour

Yeah but I was thinking about injecting incorrect "shift" in expression "value << shift".

If "value" is incorrect, then "value << shift" will be as bad as "value". However, if you anticipate that "value << shift" will be zero in case of invalid (too large) "shift", then your error processing is incorrect. Anyway, let's see if fuzzing will help.

Issue #725, opened 2021-02-10
[Q] Possible to delay scrambling?


garethfenn:
Hi there,

Is it possible to delay the scrambler plugin from starting immediately at the beginning of the stream? In other words, generate a short period (~10 seconds) of clear content allowing ECM's to be captured, ready the first key period?

Thanks for writing a great tool!

lelegard:
Hi @garethfenn 

Currently, the scrambler plugin works in "paranoid" mode exclusively: it does not let any clear packet pass. As long as the PID's to scramble are not yet known (before the first PMT of the service is received), all packets are nullified. The receiver cannot see any packet until it finishes to process the first ECM.

To do what you want, the plugin should include a new "friendly" mode where packets are passed in the clear until it is reasonable to switch to encrypted mode. Not currently implemented.

The adequate delay is normally the _transition_delay_start_ which is sent by the ECMG. It is usually much shorter that 10 seconds. 2 or 3 seconds are more common. Considering that there are 5 or 10 ECM's per second and the processing time of an ECM is less than one second (even on slow smartcards), 2 seconds _transition_delay_start_ is more than enough is most cases, even if you miss one or two ECM's.

garethfenn:
Hi @lelegard,

Thanks for your answer.

> The adequate delay...
Agreed. For live use cases this isn't required. I believe the problem in local file based playback, is that the initial content can be consumed very quickly whilst filling playback buffers... before we've had time to capture an ECM and process it. Content plays back fine from that point on wards with the default TSDuck parameters.

There are a few workarounds at the player level we could employ (or stitch some content together perhaps).... until a "friendly" mode is supported by the scrambler plugin.

lelegard:
Offline file processing is different indeed. A few years ago, I did exactly this using a shell script for a VoD PoC (in TS mode). The idea was to add null packets at the beginning of the file, enough for the initial ECM acquisition. During the initial period, PAT and PMT are cycled. Quite naturally, ECM's will also be inserted there. When the actual content starts, the player had the time to analyze the ECM's. But the idea was still a paranoid mode : make sure that the receiver is ready to descramble the first content packet but do not let any clear content.

Because the file could be quite large, the file was processed only once. There was a short initial analysis phase of the beginning of the file (get one PAT, one PMT, compute bitrate) and then, restarting at the beginning of the file, everything was done in one pass.

lelegard:
[...]

PR #1428, opened 2024-01-26
Added --clear-period to scrambler plugin

garethfenn:
#### Related issue (if any):
This is a follow up to #725, where by it is preferable to add an initial clear period of content prior to scrambling. Rationale being that there is always a delay between processing ECMs and updating the control word in the descrambler. In the first few seconds of playback commencing, before buffers preceding the decoder are not saturated, the input data is processed much faster than realtime. The first few packets are not descrambled correctly, and at least the first GOP is lost.

#### Affected components:
Tested using:
./tsp -v --add-input-stuffing 1/32 -I file ~/path/to/input.ts -P scrambler 0x01 --clear-period 10 --no-entropy-reduction --ecmg localhost:2222 --super-cas-id 0xFFFEFFFF -O file ~/path/to/output.ts

With this initial clear period, the playback path buffers can become saturated with the clear content, and allow ECMs to read and control word set ready for the first crypto period after the initial clear period. With this feature, we were able to demonstrate CAS agnostic, clear key playback of content on our hardware.

### DIFF
--- a/src/tsplugins/tsplugin_scrambler.cpp
+++ b/src/tsplugins/tsplugin_scrambler.cpp
@@ -152,6 +152,7 @@ namespace ts {
         BitRate           _ecm_bitrate = 0;             // ECM PID's bitrate
         PID               _ecm_pid = PID_NULL;          // PID for ECM
         PacketCounter     _partial_scrambling = 0;      // Do not scramble all packets if > 1
+        ts::deciseconds   _clear_period {0};            // Clear period before scrambling commences
         ECMGClientArgs    _ecmg_args {};                // Parameters for ECMG client
         tlv::Logger       _logger {Severity::Debug, tsp};  // Message logger for ECMG <=> SCS protocol
         ecmgscs::Protocol      _ecmgscs {};                // ECMG <=> SCS protocol instance.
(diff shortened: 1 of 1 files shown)


garethfenn:
Hi @lelegard,

Yes, apologies. The continuous integration checks highlighted a failure (test-019.tsp.packet.scrambler.help) on Friday, which I was only able to get around to checking on Monday. I had accidentally added the use of '-d' as a shortcut alias. However I see you have already corrected/removed it in: 41cc138f766f0329697b9378d7b8b81342aa6a45 

Thanks for integrating this change!

lelegard:
Test 019 verifies the help output of all commands and plugins. It serves two purposes:
- Detect if the help system is broken.
- Detect new or modified options, meaning the doc must be updated.

As contributors, if your contribution may affect the test suite, you may fork the `tsduck-test` repo. During the CI of a PR, if the submitter as his own `tsduck-test` fork, it is used. Otherwise, the general one, `tsduck/tsduck-test`, is used. For complicated updates, the contributor can propose a double PR, one in `tsduck` and the corresponding one in `tsduck-test`.

In the case of a simple new option, as yours, don't worry, I can handle it.

I close this PR. It won't be marked as merged, but your original commit is already merged.

PR #1465, opened 2024-05-23
Ensure python correctly interprets path specifier

paulhiggs:
It appears that recent versions of python, i.e. 3.12.3, are interpreting "\.." in the path as an escape character. 

See https://stackoverflow.com/questions/52335970/how-to-fix-syntaxwarning-invalid-escape-sequence-in-python


lelegard:
Hi Paul,

Thanks for the fix. This is clearly a bug in my script. In a Python string, a literal backslash must be doubled `'\\'`. Because `'\.'` is not a valid escape sequence, unlike `'\n'` or `'\t'` for instance, older Python versions were kind enough (or careless enough) to keep the two characters, the backslash and the dot. Python 12 correctly rejects this invalid escape sequence. I will merge your fix.

Issue #1145, opened 2022-12-16
[BUG] Physical Layer Scrambling (PLS) according to DVB API doesn't work


basicmaster:
#### Bug description:
Physical Layer Scrambling (PLS) cannot be used with the Digital Devices SX8, for the following reason:

The current implementation (as a result of #277) uses a triplet value for the `DTV_STREAM_ID` property, to provide the desired multistream ID as well as a PLS code. This seems to be [an older way to do it](https://old.tvheadend.org/issues/5556) (dvblast [does the same](https://code.videolan.org/videolan/dvblast/-/blob/master/dvblast.c#L1281)) which should probably be kept for backwards compatibility.

Meanwhile the DVB API [provides](https://git.linuxtv.org/media_tree.git/commit/include/uapi/linux/dvb/frontend.h?id=f9d79126195374c285035777b9d6abd24ceba363) the separate property `DTV_SCRAMBLING_SEQUENCE_INDEX` to set a gold code (which could be calculated from a root code, if a root code is used instead). This way is used by the mentioned card.

#### How to reproduce:
Example command for [a RAI transponder on 5°W that uses multistream and PLS](https://www.lyngsat.com/muxes/Eutelsat-5-West-B_Transalpine_11013-V.html):
```
tsp -v -I dvb -a 16 -f 11013000000 --polarity vertical --delivery-system DVB-S2 -m 8-PSK -s 35300000 --fec-inner 2/3  --pls-mode GOLD --pls-code 131070 --isi 5 -O drop
```
TSDuck doesn't lock on this transponder.

The card vendor provides [a small test tool](https://github.com/DigitalDevices/dddvb/blob/master/lib/ddzap.c) which allows to successfully lock to this transponder and to output the TS. It uses `DTV_SCRAMBLING_SEQUENCE_INDEX` for the PLS code.

#### Expected behavior:
TSDuck is able to lock on this transponder.

#### Errors and logs:
```
* dvb: using /dev/dvb/adapter16 (DVB-S, DVB-S2)
* dvb: tuned to transponder --delivery-system DVB-S2 --frequency 11,013,000,000 --symbol-rate 35,300,000 --fec-inner 2/3 --polarity vertical --modulation 8-PSK --pilots off --roll-off 0.35 --isi 5 --pls-code 131070 --pls-mode GOLD --lnb Extended --satellite-number 0
* dvb: expected bitrate from tuning parameters: 69,916,459 b/s
[...]

PR #1504, opened 2024-08-26
Added --pls-dvbapi to input plugin dvb

basicmaster:
#### Related issue (if any):
#1145

#### Affected components:
`dvb` input plugin

Adds the ability to set the PLS (GOLD) code using designated DVB API property instead of the older way to merge into the "stream id". I successfully tested this with a Digital Devices SX8 and transponders on 5.0°W and 30.0°W which use PLS.

As the `CHANGELOG.TXT` also contains a version/commit number, I didn't modify it.

I require GOLD mode to be used, as DVB API does.

When `DTV_SCRAMBLING_SEQUENCE_INDEX` is not available, the code falls back to the older way. I'm not sure, if it would be better to fail instead during parameter check...

### DIFF
--- a/doc/user/plugins/dvb-input.adoc
+++ b/doc/user/plugins/dvb-input.adoc
@@ -236,7 +236,7 @@ Must be one of the following values:
 
 |DVB-S2
 |DVB-S2
-|`--fec-inner --frequency --isi --modulation --pilots --pls-code --pls-mode --polarity --roll-off
+|`--fec-inner --frequency --isi --modulation --pilots --pls-code --pls-mode --pls-dvbapi --polarity --roll-off
  --satellite-number --spectral-inversion --symbol-rate`
 
 |DVB-T
@@ -497,7 +497,7 @@ Used for DVB-S2 tuners only.
 
 [.optdoc]
 Specify the Input Stream Id (ISI) number to select, from 0 to 255.
-Used with multi-stream, see also options `--pls-code` and `--pls-mode`.
+Used with multi-stream, see also options `--pls-code`, `--pls-mode` and `--pls-dvbapi`.
 
 [.optdoc]
 The default is to keep the entire stream, without multi-stream selection.
@@ -591,6 +591,28 @@ Specify the Physical Layer Scrambling (PLS) mode.
 Used with multi-stream, see also option `--isi`.
 Must be one of `COMBO`, `GOLD`, `ROOT`.
 The default is `ROOT`.
+
+[.optdoc]
+*Warning*: this option is supported on Linux only.
+Currently, Windows provides no support for multi-stream.
+
+[.opt]
+*--pls-dvbapi*
+
+[.optdoc]
+Used for DVB-S2 tuners only.
+
+[.optdoc]
+When specified, the PLS code is set through the designated DVB API property
+instead of the older way to merge into the "stream id".
+Used with multi-stream, see also option `--isi`.
+
+[.optdoc]
+If this parameter is used, the PLS GOLD mode must be used.
+
+[.optdoc]
+The default is to merge the (left shifted) PLS code and PLS mode into the `DTV_STREAM_ID` property.
+DVB API later added the `DTV_SCRAMBLING_SEQUENCE_INDEX` property as standardized way to specify a PLS (GOLD) code.
 
 [.optdoc]
 *Warning*: this option is supported on Linux only.
--- a/src/libtsduck/dtv/broadcast/tsModulationArgs.cpp
+++ b/src/libtsduck/dtv/broadcast/tsModulationArgs.cpp
@@ -48,6 +48,7 @@ void ts::ModulationArgs::clear()
     isi.reset();
     pls_code.reset();
     pls_mode.reset();
+    pls_dvbapi.reset();
     sound_broadcasting.reset();
     sb_subchannel_id.reset();
     sb_segment_count.reset();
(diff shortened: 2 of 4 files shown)


lelegard:
Hi @basicmaster 

Thanks for the contribution.

There are a few things I don't fully understand:

- When `DTV_SCRAMBLING_SEQUENCE_INDEX` is not defined (older kernel), the PLS code is set in the id. Does it work? You mentioned in #1145 that you cannot lock the signal. So, why do everyone use this method?

- When `DTV_SCRAMBLING_SEQUENCE_INDEX` is defined (newer kernel), why do we need to defined a new option `--pls-dvbapi`? If the new method is to set the PLS code in this option, let's do it unconditionally. If this is the way to do, there is no need to put the burden on the user to add an option. Especially because no one will notice that this option is needed on their version of the kernel. So, in practice, the option will not be used and the command will not work.

When the system API changes, the application should detect it and apply it but this should remain transparent to the user.

lelegard:
Hi @basicmaster 

I tried to extract your code, without the command line option. It pushed that in a temporary branch named `pls-code`, in commit https://github.com/tsduck/tsduck/commit/58aa48ccf7a6a887ccfc9bf433e02a4dbcb7ddd4

The changes are minimal. Could you check if this work for you on your streams with PLS code?

There is one thing, however, which must be wrong in that code. When `DTV_SCRAMBLING_SEQUENCE_INDEX`, we assume that the PLS mode is GOLD, which is incorrect.

In the old API, the PLS mode and code were included in the stream ID. On reading the parameters, we extracted them from the id, including the _PLS mode_. Here, we set it to always to GOLD. Do you know how we can extract the PLS mode from the new API? And depending on that mode, which PLS code shall we set?

basicmaster:
> There are a few things I don't fully understand:
> 
> * When `DTV_SCRAMBLING_SEQUENCE_INDEX` is not defined (older kernel), the PLS code is set in the id. Does it work? You mentioned in [[BUG] Physical Layer Scrambling (PLS) according to DVB API doesn't work #1145](https://github.com/tsduck/tsduck/issues/1145) that you cannot lock the signal. So, why do everyone use this method?

The card I use doesn't support that older way - thus I cannot tell whether it works. But I think so, as otherwise someone had already complained.

> * When `DTV_SCRAMBLING_SEQUENCE_INDEX` is defined (newer kernel), why do we need to defined a new option `--pls-dvbapi`? If the new method is to set the PLS code in this option, let's do it unconditionally. If this is the way to do, there is no need to put the burden on the user to add an option. Especially because no one will notice that this option is needed on their version of the kernel. So, in practice, the option will not be used and the command will not work.
> 
> 
> When the system API changes, the application should detect it and apply it but this should remain transparent to the user.

My intention to add the parameter was to allow switching between old/current variant, so people can still also use ROOT and COMBO codes. But I think it is fine to abstain from adding the parameter and to just use the current API variant, if available.

---

Your `pls_test` branch works as expected: I'm able to tune different DVB-S2 transponders with and without PLS. In case of PLS, I just use GOLD codes, together with different multistream IDs.

---

The "new" property of the DVB API actually [expects a GOLD code to be specified](https://git.linuxtv.org/media_tree.git/commit/include/uapi/linux/dvb/frontend.h?id=f9d79126195374c285035777b9d6abd24ceba363) - so there is no API field to specify which mode of GOLD/ROOT/COMBO to use. This is why I added a PLS mode parameter value check in my branch and why I set the mode to GOLD when the current PLS code is retrieved from the card.

Digital Devices did a deeper elaboration on the different PLS code modes in a long comment block of their PLS code conversion tool (the tool [meanwhile has been deleted during a clean up](https://github.com/DigitalDevices/dddvb/commit/c0e174505a59a5bf644939b4f391cf225a7acb01) unfortunately):
https://github.com/DigitalDevices/dddvb/blob/87478ec63abde8817eb4519811d9969727b34850/apps/pls.c#L38

I understand that GOLD code and ROOT code have a specific mathematical relation to each other i.e. for a particular GOLD code there is a corresponding particular ROOT code that has the same effect. According to the comment block, COMBO mode doesn't make (anymore) sense.

So if ROOT code shall continue to be supported (I myself don't need it, but others may), it would make sense to include a similar conversion as in the tool from ROOT to GOLD, before handing over the PLS code to DVB API, as it expects GOLD. The tool's conversion is actually rather doing brute force; I don't know whether there is a more elegant approach or whether brute-force is the only way to do this.

As described in the mentioned code comment, COMBO could/should be removed then. At least it won't work with the current DVB API variant. Lyngsat also doesn't seem to have anymore entries with COMBO code.

lelegard:
Hi @basicmaster 

Thanks for the details, the feedback and the links.

I merged branch `pls-mode` into `master` and did the following changes:

- COMBO mode is now removed. You are right, it seems a legacy transient buggy feature.
- GOLD mode is now the default. This is the one which is documented by ETSI EN 302 307. It should be the reference one. The default PLS code is zero.
- When ROOT mode is specified, the PLS code must be specified. Previously, the default mode was ROOT and the default code was zero, which is meaningless since GOLD code 0 (the default) means ROOT code 1.
- With `DTV_SCRAMBLING_SEQUENCE_INDEX`, when the PLS mode is specified as ROOT, the PLS code is converted to a GOLD one before setting the parameter in the DVB API.

I think that this makes more sense.

Could you please try the latest state of the repo (branch `master`) and let us know if it works correctly for you?

basicmaster:
I agree it makes sense to make GOLD the default mode, as it is the actual code to work with.

I can confirm that the current `master` branch works as expected. Using `grep` I just found [one line in the docs that has to be updated yet](https://github.com/tsduck/tsduck/blob/master/doc/user/20B-app-channel-config.adoc?plain=1#L126).

Thank you for working this out, and for the additions and improvements. This looks good to me now.

lelegard:
> Using grep I just found [one line in the docs that has to be updated yet](https://github.com/tsduck/tsduck/blob/master/doc/user/20B-app-channel-config.adoc?plain=1#L126).

Thanks, fixed.

PR #1507, opened 2024-08-29
Add AVS audio descriptors from T/AI 109.7

paulhiggs:
These descriptors indicate the characteristics of AVS2 and AVS3 coded audio in an elementary stream. They are defined in T/AI 109/7

### DIFF
--- a/src/libtsduck/base/cpp/tsPlatform.h
+++ b/src/libtsduck/base/cpp/tsPlatform.h
@@ -860,7 +860,6 @@ TS_MSC_NOWARNING(5045)  // Compiler will insert Spectre mitigation for memory lo
 #include <cstddef>
 #include <climits>
 #include <cmath>
-#include <variant>
 #include <fcntl.h>
 #if defined(TS_WINDOWS)
     #include <windows.h>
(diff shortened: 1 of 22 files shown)


lelegard:
Hi Paul,

Thanks for the update. Clearly, you are the only one in position to provide this!

One question: Why did AVS create two PDS, one for AVS video and one for AVS Audio? Usually, a PDS is allocated to an organization or company. There is one single PDS for UK DTG, EU EACEM, BSkyB, or Eutelsat. The PDS owner manages its own allocation of descriptor tags.

Are there two distinct standard bodies, AVS video and AVS audio, which want to be independent?

Given that all the current 3 AVS-defined descriptors have distinct tags, (D1, D2, D3), why did they create two distinct PDS? In addition to the fact that they have to pay DVB for two PDS instead of one, there is no technical necessity.

I see that the initial one, 0x41565356 for video, is [officially registered in DVB services](https://www.dvbservices.com/identifiers/private_data_spec_id). I suppose that the new one, 0x41565341 for audio, is not yet published.

lelegard:
Merged. I added AVS2Audio in CodecType as well.

PR #1513, opened 2024-09-04
Small updates related to AVS codecs and codec types

paulhiggs:
Couple of updates for AVS2/AVS3 codecs relate to checking content type
Update to use a set lookup, rather than a large switch statement to determine Audio/Video/Subtitling


lelegard:
Thanks Paul. Definitely a good idea. I just moved the constant sets to the static anonymous namespace to avoid building them each time a CodecTypeIsXXX function is called.

PR #1524, opened 2024-09-25
tsplugin_mux update

nabilervatra:
#### Related issue (if any):

The current mux plugin does not generate transport stream with accurate bitrate.
#### Affected components:

The only change I made is on `tsplugin_mux.cpp`.
I set the `_pid_next_pkt` to be more accurate by calculate it on the fly. Rather than fixing an `_inter_pkt` that pre determined and truncated to the nearest integer, so It will make the resulted bitrate far from the target bitrate.


lelegard:
Thanks. The `mux` plugin is an old and would deserve a lot of work...

PR #1530, opened 2024-10-03
Fix Windows build script

lgu-appear:
#### Affected components:
No TSDuck core code is affected since it is only about the installation files for Windows.

Add --accept-source-agreements flag to remove user interaction which locked the script and could not run farther. See https://learn.microsoft.com/en-us/windows/package-manager/winget/install#license-agreements

However the flag they propose is discontinued and the one I added is the one that works. Manage to try it on another Windows computer that did not agree to the agreements yet using winget.

### DIFF
--- a/scripts/install-wintools/install-ruby.ps1
+++ b/scripts/install-wintools/install-ruby.ps1
@@ -29,7 +29,7 @@ else {
     # Ruby is unconventional here. Instead of a constant package name and different
     # versions, the package name contains the version. Each version has a new package
     # name. Get latest version.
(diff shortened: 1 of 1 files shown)


lelegard:
Thank you for the fix.

FYI, the latest state of the repo is rebuilt every night and the Windows and Ubuntu binaries are available at https://tsduck.io/download/prerelease/

If you rebuilt TSDuck on Windows just because you needed the latest version, grabbing the latest nightly builds saves time. Of course, you are welcome to rebuild it for any other purpose, such as developing new code or simply technical curiosity.

PR #1534, opened 2024-10-11
Fixed the issue where Linux could not handle multiple devices.

a83912a:
* Fixed the issue where Linux could not handle multiple devices.
* Windows devices are currently unable to support multiple devices (this will be fixed libvatek in the next version update).

### DIFF
--- a/src/libtsduck/vatek/tsVatekControl.cpp
+++ b/src/libtsduck/vatek/tsVatekControl.cpp
@@ -35,7 +35,7 @@ ts::VatekControl::VatekControl(int argc, char *argv[]) :
     analyze(argc, argv);
 
     const bool all = present(u"all");
-    getIntValue(_dev_index, u"", all ? -1 : 0);
+    getIntValue(_dev_index, u"", all ? -100 : 0);
     if (all && _dev_index >= 0) {
         error(u"specifying a device index and --all are mutually exclusive");
     }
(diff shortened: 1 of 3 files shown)


lelegard:
@a83912a, this does not work. The call to `vatek_device_list_enum_multiple` should be conditioned to `#if VATEK_VERSION >= ...`. You can see in the log that the macOS builds fail. The latest libvatek version in Homebrew is still 3.10 and the compilation fails.

See all my comment here: https://github.com/VisionAdvanceTechnologyInc/vatek_sdk_2/issues/9#issuecomment-2407244457

The build of applications (such as TSDuck) is broken on Windows.

Please revert asap the latest release to 3.10. It the autobump job of Homebrew runs before that, it will grab your 3.12 version and propagate the problems on all macOS systems.

Please perform more thorough testing on all platforms before publishing a release. This means build a small application using libvatek (no need to build the TSDuck beast) on all operating systems.

lelegard:
I suggest that you add your own CI job in repository VisionAdvanceTechnologyInc/vatek_sdk_2, just like TSDuck does.

The CI job should run on Linux, Windows, macOS:
- Build libvatek
- Install libvatek
- Build a small application which only calls `vatek_device_list_enum` and/or `vatek_device_list_enum_multiple` (no VATek device is expected on the CI runner system). See [example from Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/83aff977ba2b18330a0ca177feb3ded4f43392ae/Formula/lib/libvatek.rb#L30).
- Run that small application and check that no error is reported.

With that kind of job, all tests are run everywhere each time you push a commit. If you develop and test on platform A but your modifications create a problem on platform B, you can see it right after the push. This is the way TSDuck works and this is invaluable. It regularly happens that something I develop on macOS fails on Windows or (more rarely) on Linux. I am notified within minutes that something is wrong with the latest commits.

lelegard:
Closed. New PR expected with a more stable libvatek (see https://github.com/VisionAdvanceTechnologyInc/vatek_sdk_2/issues/9#issuecomment-2408622735)

PR #1548, opened 2024-11-14
Updates related to NGA component_descriptors

paulhiggs:
Adding component_descriptor values for DTS-UHD
Add parsing of NGA component_descriptor tag values
Fix formatting to avoid line wrapping


lelegard:
Thank you Paul

Issue #1549, opened 2024-11-16
[BUG] Buiild failure with v3.39-3956


daeho-ro:
#### Bug description:
During build, we got an error,
```
  ==> gmake NOGITHUB=1 NOTEST=1
  /private/tmp/tsduck-20241116-13602-zpsojw/tsduck-3.39-3956/scripts/make-config.sh: line 406: unexpected argument `]]' to conditional unary operator
  /private/tmp/tsduck-20241116-13602-zpsojw/tsduck-3.39-3956/scripts/make-config.sh: line 406: syntax error near `;'
  /private/tmp/tsduck-20241116-13602-zpsojw/tsduck-3.39-3956/scripts/make-config.sh: line 406: `if [[ -z$CROSS$CROSS_TARGET ]]; then'
  bash: /git-hook-update.py: No such file or directory
  gmake: *** [Makefile:106: git-hooks] Error 127
```
and suspected line here,
- https://github.com/tsduck/tsduck/blob/a18add90bf2e34ce1ff54ba357fc0ae88721368c/scripts/make-config.sh#L406

#### How to reproduce:
Download source, run command that is all I guess.
```
gmake NOGITHUB=1 NOTEST=1
```

#### Expected behavior:
Should build without error

#### Errors and logs:
As uploaded above
~~~
[be sure to surround verbatim log lines with "~~~" lines]
~~~

#### Environment:
 - OS: macOS
 - OS version: 14.7 arm64
 - TSDuck full version: 3.39-3956 maybe
 - Installation type: own build

PR #1550, opened 2024-11-16
fix syntax issue in make-config.sh

chenrui333:
#### Related issue (if any):

fixes https://github.com/tsduck/tsduck/issues/1549

relates to https://github.com/Homebrew/homebrew-core/pull/197908


lelegard:
Thank you.

Unique behaviour when building in a Homebrew environment. See my comment here: https://github.com/Homebrew/homebrew-core/pull/197908#issuecomment-2480631264

chenrui333:
Thanks @lelegard!

PR #1559, opened 2024-12-01
Updates for timing and HRD descriptors

paulhiggs:
Consistently handled MPEG defined timing and HRD descriptors for AVC, HEVC, EVC and VVC.
In the descriptor definitions in ISO/IEC 13818-1, the **90kHz_flag** indicates indicates that the frequency of the time base is 90 kHz. The **N** and **K** values are used to define the time scale for other frequencies.


lelegard:
Much more clear. Thank you.

PR #1562, opened 2024-12-16
HDR Vivid descriptor and AVS2 audio descriptor are only in MPEG context (no DVB PDS)

paulhiggs:
Also, build registration ids from 'readable' values.


lelegard:
Good idea this `MakeREGID`.

PR #1572, opened 2025-01-06
be more like c++ strings with starts_with and ends_with

paulhiggs:
c++ 20 defines [`starts_with`](https://en.cppreference.com/w/cpp/string/basic_string/starts_with) and [`ends_with`](https://en.cppreference.com/w/cpp/string/basic_string/ends_with) so use these with UString for consistent readability.

### DIFF
--- a/sample/sample-java/.project
+++ b/sample/sample-java/.project
@@ -1,17 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-  <name>sample-java</name>
-  <comment></comment>
-  <projects>
-  </projects>
-  <buildSpec>
-    <buildCommand>
-      <name>org.eclipse.jdt.core.javabuilder</name>
-      <arguments>
-      </arguments>
-    </buildCommand>
-  </buildSpec>
-  <natures>
-    <nature>org.eclipse.jdt.core.javanature</nature>
-  </natures>
+	<name>sample-java</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+	</natures>
+	<filteredResources>
+		<filter>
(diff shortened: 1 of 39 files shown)


paulhiggs:
Generally, overloading functions with non-ambiquous arguments is an improving nature of modern languages and removing the human decision on whether to call starts_with or startsWith seems like a good thing. Ideally, if UString::starts_with is called with default arguments, the rhe result of string::starts_with should be returned. Only is "UString enhanced" functions are needed should the specific behaviour be coded.

My PR doesn't have all that - but could do..

paulhiggs:
There may be a bounds issue with UString::startsWith and UString::endsWith if the length of the prefix/suffix is greater than the string being evaluated, i.e. `UString("paul").startsWith(u"thierry")` would likly crash as the range check is performed using the length of the argument to evaluate

lelegard:
Alright. In that case, we should provide these to be a strict superset:
~~~
bool startsWith(const std::u16string_view& prefix, CaseSensitivity cs = CASE_SENSITIVE, bool skip_spaces = false, size_type start = 0) const;
bool startsWith(UChar prefix, CaseSensitivity cs = CASE_SENSITIVE, bool skip_spaces = false, size_type start = 0) const;
bool startsWith(const UChar* prefix, CaseSensitivity cs = CASE_SENSITIVE, bool skip_spaces = false, size_type start = 0) const;

bool endsWith(const std::u16string_view& suffix, CaseSensitivity cs = CASE_SENSITIVE, bool skip_spaces = false, size_type end = NPOS) const;
bool endsWith(UChar suffix, CaseSensitivity cs = CASE_SENSITIVE, bool skip_spaces = false, size_type end = NPOS) const;
bool endsWith(const UChar* suffix, CaseSensitivity cs = CASE_SENSITIVE, bool skip_spaces = false, size_type end = NPOS) const;
~~~

lelegard:
Paul, for some reason, in one of your commits, the files `.project` were overwritten in at least two java directories. They were reformatted and contain lines such as this:
~~~
<arguments>node_modules|\.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
~~~
Some Visual Studio configurations on your side maybe? Or another processing tool?

These files are used by Eclipse when working on the Java files and should not change.

PR #1575, opened 2025-01-13
Draft: ETSI TR 101-290 Analyzer

### LINKED ISSUE #559: TR 101 290 analysis
Hello, 

Do you plan to add a TR 101 290 analysis to TsDuck in the future ?
(full analysis that includes priorities 1/2/3 monitoring)

Best regards.

### LINKED ISSUE #1267: Is there a way to monitor TR 101 290 errors using a API library?

stazio:
Hello!
This is my first pass at implementing a [TR 101-290 (2020-06)](https://www.etsi.org/deliver/etsi_tr/101200_101299/101290/01.04.01_60/tr_101290v010401p.pdf) analyzer. I based my work off of the `analyze` plugin, and some of the PCR calculations from `pcrverify`. I would love if someone could take a look and see if my usage of `mdata.getInputTimeStamp()` is reasonable, and if there are any other obvious issues. 

The biggest issue right now is how the analysis is printed, as it just dumps texts directly to the console without taking advantage of any of the pretty printers. Opinions on the best way to stylize the output would be much appreciated!

Also, I need some help figuring out a better way to interact with the `SectionDemux` class. Currently, it is a little bit vague when it comes to providing error feedback (especially on a per-pid level, such as CRC errors). The specification is a little bit vague here as well, but I think that ideally, we should increment `PAT_error` when non-PSI data is transmitted on a PAT/PMT pid (or when those packets are corrupted). Additionally, I don't see a way to access the base timestamp of the underlying packets, which currently makes it impossible to validate that the PAT/PMT occurs frequently enough.

Please be aware that it is my intention to rebase this PR to have a clean history once development finishes up! I will avoid force pushing...for now.

TODO:
- [ ] Refactor how reports are printed to take advantage of TSduck's pretty printers.
- [ ] Add support for logging errors immediately, so you can print an error instantly (like it is done in `pcrverify`), rather than on an interval.
- [ ] Add support for dumping reports to JSON/XML.
- [ ] Investigate if there are ways to report CRC errors per-PID, rather than globally.
- [ ] Consider adding a lazy checker to make sure that if a PID stops forever, we still print the error.
- [ ] For errors where this makes sense, report the min/max/current value, such that if there are, for example, PCR accuracy errors, we can see how inaccurate they really are.
- [ ] Cull PIDs that do not show up for a long period of time. We will report on every PID in a stream, regardless of if the PAT/PMT actually assigns itself to one.
- [ ] Add priority 3 errors.

#### Related issue (if any):
Fixes #559.
Fixes #1267.

#### Affected components:
Introduces a new TSP processor called `tr101_290`.


lelegard:
Hello @stazio,

Thanks a lot for this new feature. Could you have a look at the CI failures? You may have a look at other classes in TSDuck for coding styles and practices.

lelegard:
> Also, I need some help figuring out a better way to interact with the SectionDemux class. Currently, it is a little bit vague when it comes to providing error feedback (especially on a per-pid level, such as CRC errors).

I am not sure to understand what you mean about error feedback.

- All message reporting are done through instances of `ts::Report`. This is an abstract class which is derived in many ways.
- The interpretation of tables is also subject to a context which is maintained by an instance `ts::DuckContext` which is usually created by the application. You can access the associated report through the method `ts::DuckContext::report()`.
- In the case of a plugin, there is one `ts::DuckContext` per plugin. It is created by `tsp` and passed to the plugin as protected field `duck`.
- From inside a plugin class, message reporting is direct since `ts::Plugin` is an [indirect subclass if `Report`](https://tsduck.io/doxy/d5/d8a/classts_1_1ProcessorPlugin.html).

Also note that the class `ts::TSP` is a proxy between `tsp` and its plugins. It should not be used in the general-purpose classes of the library (as used in the constructor of your `TR101_290Analyzer`). This would prevent to use that class in a generic application, outside a plugin. If a general-purpose class does packet processing, I would recommend to pass a `ts::DuckContext` in the constructor and keep a reference to it.

> I don't see a way to access the base timestamp of the underlying packets, which currently makes it impossible to validate that the PAT/PMT occurs frequently enough.

What do you mean by "base timestamp"? The reception timestamp of a packet (from the input plugin) is part of the associated packet metadata, the instance of `ts::TSPacketMetadata` which is passed to the `processPacket()` method of the plugin.

> Please be aware that it is my intention to rebase this PR to have a clean history once development finishes up! I will avoid force pushing...for now.

So, you mean that this PR is just for review and you don't expect a merge now, isn't it?

lelegard:
Also please do not use `std::string`, these are 8-bit strings which not are suitable for international information. TSDuck uses `ts::UString` everywhere, 16-bit strings with additional features.

stazio:
So the issues I have had so far are in regards to `SectionDemux`. 

1. I am not sure how exactly I could identify in a Report on which PID a particular CRC error occurred. Thinking about it now, I supposed I could make a new `SectionDemux` instance for each PID instead and treat them that way.
2. I don't have access to `mdata` inside of `handleTable`. I'm trying to get the timestamp of the last packet that generated each table, to confirm that they occur frequently enough. I just pushed [4579140](https://github.com/tsduck/tsduck/pull/1575/commits/4579140d8ada24ccd27dda37371df53c09377912) which works around the issue, but I am not sure if that is correct. I am just storing the latest timestamp I have received as a variable before pushing it into the demuxer, and then using that variable within `handleTable`.
3. Is there a way to have `SectionDemux` notify me on a repeated table (with the same `version` field). Part of the spec is to validate that the PAT and PMT occurs frequently enough, and right now I am only getting the first PAT/PMT in a TS, and no repeats of it.

By the way, this is a very lovely code base to be working in. It has been super easy to just jump in and write new stuff!

lelegard:
> I am not sure how exactly I could identify in a Report on which PID a particular CRC error occurred. Thinking about it now, I supposed I could make a new SectionDemux instance for each PID instead and treat them that way.

No need to have many demux. The classes `DemuxedData`, `Section` and `BinaryTable` have a `sourcePID()` method.

> I don't have access to mdata inside of handleTable. I'm trying to get the timestamp of the last packet that generated each table, to confirm that they occur frequently enough.

In your analyzer class, you may have a `TSPacketMetadata` private field. Each time you receive a packet, before sending it to the demux, copy the `TSPacketMetadata` of the packet into that field. Assuming that your handler virtual methods for section or table are in the same class, they are access to that field. And these handlers are called on the last packet that triggered the creation of the section. Therefore, you have the time stamp of the last packet of the section.

> Is there a way to have SectionDemux notify me on a repeated table (with the same version field). Part of the spec is to validate that the PAT and PMT occurs frequently enough, and right now I am only getting the first PAT/PMT in a TS, and no repeats of it.

Use a section handler, not a table handler. A section handler is called for every section, all the time, while a table handler is called only on a new version of a table (for long tables). But you have to manage section numbers since the handler is called for every section of every table.

lelegard:
Which operating system do you use? It seems that the CI fail on all operating systems and all compilers. This means it wouldn't build on your system. Does it?

stazio:
> Which operating system do you use? It seems that the CI fail on all operating systems and all compilers. This means it wouldn't build on your system. Does it?

I am compiling using my meson nonsense, so I probably don't have `-Werror` set on that. I am on Linux. I'll try and get it compiling later today.

> No need to have many demux. The classes DemuxedData, Section and BinaryTable have a sourcePID() method.

Does the callback still get called, even on a failed CRC check?

> Use a section handler, not a table handler. A section handler is called for every section, all the time, while a table handler is called only on a new version of a table (for long tables). But you have to manage section numbers since the handler is called for every section of every table.

Okay. I'll look into this.

lelegard:
> I am compiling using my meson nonsense, so I probably don't have -Werror set on that. I am on Linux. I'll try and get it compiling later today.

Don't do that! The build system is complex, don't try to mess with it unless you want to waste hours of your time.

The build system of TSDuck is designed to be simple to use. Just add new code, new source files and `make`. Well, `make -j10` at least (for 8 cores), otherwise you wait for hours.

Compiling without warning report for days and later trying to fix them all at once is a very bad method. Learn to spot the potential bugs from the beginning. Code, compile, fix, test and iterate. That's the principle of continuous integration.

> >    No need to have many demux. The classes DemuxedData, Section and BinaryTable have a sourcePID() method.
>
> Does the callback still get called, even on a failed CRC check?

Implement a `InvalidSectionHandlerInterface`. It is called when an invalid section is found (invalid size, invalid CRC, whatever the demux doesn't like). See an example in `ts::TablesLogger::handleInvalidSection`.

stazio:
I just compiled using `make` and fixed all the build errors so the build process should succeed now.

I used meson because I wanted some Intellisense from CLion and this was the fastest way I knew how .

lelegard:
> I used meson because I wanted some Intellisense from CLion and this was the fastest way I knew how .

On macOS and Linux systems, I use Qt Creator. It uses a qmake / cmake project structure which is maintained in parallel in `scripts/qtcreator`. There is minimal setup to be allowed to compile and debug under Qt Creator but the "real" build is done from the command line using `make`.

Qt Creator is extremely powerful and free for non-commercial projects whereas CLion is not free for open source projects. It is only free for "students and teachers".

On Windows, the build system is MSBuild (some kind of XML-based make), maintained in `scripts/msvc`. The MSBuild files are directly used by Visual Studio. All tools are completely free for open source projects.

The makefiles are sufficiently generic to automatically discover new subprojects, executables and plugins but Qt Creator and MSBuild (just like CMake, don't know meson or conan) need specific files per project. The creation and consistency of new subprojects is maintained by `scripts/build-project-files.py`. This script automatically detects all projects, like the makefiles, and create missing entries in `scripts/qtcreator` and `scripts/msvc`. 

With CMake, it is even worse, because all source files need to be individually listed. This is insane in large dynamic projects without large development teams, this is a prehistoric approach. I know that techniques exist to avoid this but CMake zealots insist that this must not be used. There are some valid reasons. So I just sent them to hell...

stazio:
Hey @lelegard ,
Added JSON printing, removed my usage of std::string, and fixed more code smells (so hopefully it compiles now).

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
```diff
@@ -0,0 +1,503 @@
...
+                    it->second->_type = ServiceContext::Unassigned;
+                }
+            }
+        }
+    }
+}
+
+void ts::TR101_290Analyzer::handleSection(SectionDemux& demux, const Section& table)
+{
+    if (table.sectionNumber() != 0) {
```
stazio, line 93: I hope this is what you meant when you said that I had to look out for section numbers.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+        for (auto & [_, s2] : _services) {
+            s2->cat_error = 0;
+        }
+    }
+}
+void ts::TR101_290Analyzer::handleInvalidSection(SectionDemux& demux, const DemuxedData& data) {
+    auto _service = getService(data.sourcePID());
+    // todo: it is not guaranteed that it is a CRC error that causes this issue.
+    _service->crc_error++;
```
stazio, line 146: Am I bad in assuming that this will always be a CRC error? If not, how do I check if this was caused by a CRC error?

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
```diff
@@ -0,0 +1,503 @@
...
+{
+    if (!json.loadArgs(duck, args)) return false;
+
+    show_report = !json.useJSON() || args.present(u"show-report");
+    return true;
+}
+void ts::TR101_290Analyzer::processPacket(ServiceContext& ctx, const TSPacket& pkt, const TSPacketMetadata& mdata) const
+{
+
+    // todo: do TS packets ever arrive with invalid sync bytes this deep in?
```
stazio, line 180: I tried to test some Sync errors, but they didn't make their way all the way down into the plugin. Is there a way to query for these errors?

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
```diff
@@ -0,0 +1,503 @@
...
+ErrorState get_repetition_error(const ts::TR101_290Analyzer::ServiceContext& ctx) {return {ctx.pcr_repetition_err.count, ctx.has_pcr, ctx.pcr_repetition_err};}
+ErrorState get_discontinuity_error(const ts::TR101_290Analyzer::ServiceContext& ctx) {return {ctx.pcr_discontinuity_err.count, ctx.has_pcr, ctx.pcr_discontinuity_err};}
+ErrorState get_accuracy_error(const ts::TR101_290Analyzer::ServiceContext& ctx) {return {ctx.pcr_accuracy_err.count, ctx.has_pcr, ctx.pcr_accuracy_err};}
+ErrorState get_pts_error(const ts::TR101_290Analyzer::ServiceContext& ctx) {return {ctx.pts_err.count, is_pes(ctx), ctx.pts_err};}
+ErrorState get_cat_error(const ts::TR101_290Analyzer::ServiceContext& ctx) {return {ctx.cat_error, true};}
+
+void ts::TR101_290Analyzer::report(std::ostream& stm, int& opt, Report& rep)
+{
+    stm << "Priority 1 Errors:\n";
+    // todo: TS_sync_loss may not be a valid test in IP-based systems and covered by Sync_byte_error.
```
stazio, line 416: TS_sync_loss is defined as: 
> Loss of synchronization with consideration of hysteresis parameters

I am not entirely sure what this means, or if it even applies in file or IP-based formats. Any ideas?

stazio (commented):
Left some comments here. If you could take a look and let me know if you have any ideas on how best to solve these problems, that would be appreciated.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
lelegard: Yes. The table cycle is the distance between two sections with the same number (typically section #0), not the distance between two sections (for instance sections 0 and 1 of the same table).

lelegard: And, for the sake of readability, you should call your parameter `section`, not `table`  (not the same concept, one is a subcomponent of the other).

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
lelegard: No, there are many possible reasons for a section to be invalid.
See an example here:
https://github.com/tsduck/tsduck/blob/cc250477985e3e637ad03ae9f89bbdae6294655a/src/libtsduck/dtv/signalization/tsTablesLogger.cpp#L822

I recomputes the CRC to determine if this is a CRC error.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
lelegard: Not easy or even meaningful to test with a generic framework such as TSDuck. This is something to test at demodulator level. And even at modulator level, it is not really meaningful to differentiate corruption of that byte, compared to other byte. This is a bit artificial.

At tsp level, packets with corrupted sync byte will probably never be output by the receiver or input plugin.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
lelegard: > I am not entirely sure what this means

Same for me...

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
lelegard: Since the demux already did all checks before declaring a section invalid, it is more consistent to propagate that status to the invalid section handler, rather than letting the handler do the job again.

Therefore, I added an enum type for section status and the profile of the handler has been modified to add it as parameter:

https://github.com/tsduck/tsduck/blob/061b28307c5522d70859c34a287451eed9eba68e/src/libtsduck/dtv/demux/tsInvalidSectionHandlerInterface.h#L38

The status values are defined in class `Section`:

https://github.com/tsduck/tsduck/blob/061b28307c5522d70859c34a287451eed9eba68e/src/libtsduck/dtv/signalization/tsSection.h#L53

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
stazio: Okay I found some more information about this particular error:

> TS_sync_loss
The most important function for the evaluation of data from the MPEG-2 TS is the sync acquisition. The actual
synchronization of the TS depends on the number of correct sync bytes necessary for the device to synchronize and on the number of distorted sync bytes which the device cannot cope with. 

> It is proposed that five consecutive correct sync bytes (ISO/IEC 13818-1 [i.1], clause G.1) should be sufficient for sync acquisition, and two or more consecutive corrupted sync bytes should indicate sync loss. 

> After synchronization has been achieved the evaluation of the other parameters can be carried out.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
stazio: So, I think testing for this is still important.

 TR 101 290 defines this as:
>  The indicator "Sync_byte_error" is set as soon as the correct sync byte (0x47) does not appear after 188 or 204 bytes.

In a file this is super easy to produce by just deleting a couple of bytes from the middle of a file recording.

 I don't think this error would ever trigger in practice in an IP stream, but I could see some potential condition, where if you converted IP packets to a byte stream and then back to IP while taking loss in the byte stream, it can be possible for the conversion back to IP to cause a sync loss.

In both cases, TSDuck would be the "demodulator", especially if you're performing more complex operations, such as a `regulate`.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
stazio: Sweet! Added, thanks!

stazio:
Just FYI, I rebased and force pushed to get the `Section::Status` stuff merged into this branch.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
lelegard: OK, so it seems the same topic as the 0x47 sync bytes. As I previously commented, this is a demodulation issue, not something that can be reliably tested in TSDuck. In the case of the `dvb` input plugin, the resync is done by the demod and we start receiving packets once the demod has resync. So, we can't really measure it. This must be measured by specialised equipment, or maybe by some SDR software using a RF receiver.

stazio:
Hey @lelegard,
I am going to take the draft off of this PR, as I think we're getting pretty close to the final product! Apologies about how long this took -- I ran into a bunch of issues in my original implementation when trying to deal with timeouts and became disillusioned with the code, and I also had to step away for a couple of other reasons. After some refactors, and the introduction of the `Indicator` class, I am much more happy with this design.

Could you trigger the workflows again? I expect them to fail again but I can take a look at what is going on with them.

I left a couple of todo comments throughout the code base as it doesn't strictly follow the specification, and so I want to see if I can work those out. I also need to do a bunch of testing. We're also currently still missing Priority 3 errors, so I will see if I can get those done in this PR or make a follow-up. I also want to add some more docs.

At this point, if you wanted to take a look at the code and see if there is anything you would like to change, let me know. I have mixed opinions on how I am dealing with printing each log report (you can tell I am a C developer :joy: ), but I don't strictly have an issue with it either.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
stazio: Following up on this up in the earlier comment.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
stazio: @lelegard pining again, as I think my previous comment got missed.

stazio (commented):
Working my way through these todo's. Added support for Priority 3 errors, and cleaned up the real time error prints. 

Biggest challenge at this point would be to support the Priority 3 buffer error conditions. I want to consider doing this in a separate PR, as I have stayed well clear of dealing with buffer errors as I know they are complicated in TSes.

I am planning on testing this with some real world feeds this weekend.

src/libtsduck/dtv/analysis/tsTr101Analyzer.cpp
```diff
@@ -0,0 +1,972 @@
...
+                info(*service, service->RST_error, u"two sections with table_id=0x70 (TDT) occurred too frequently: %d", currentTimestamp - service->last_table_actual);
+
+            service->last_table_actual = currentTimestamp;
+            service->last_table = currentTimestamp;
+        }
+    } else if (table.sourcePID() == TID_CAT) {
+        auto cat = CAT(_duck, table);
+
+        for (auto it : cat.descs.list()) {
+            auto ca = CADescriptor(_duck, *it);
```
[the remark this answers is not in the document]
lelegard: Last night, I pushed a commit on the master branch with iterators on `DescriptorList`. These iterators give access to the binary `Descriptor` objects.

Additionally, the index `[]` operator on `DescriptorList` has been redefined to give access to the binary `Descriptor` object instead of the shared pointer to it. Since these pointers are never null by design of the `DescriptorList` class, this eliminates a level of indirection and avoids the well-justified temptation to test the pointer against the null value.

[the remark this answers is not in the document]
stazio: Updated, thanks!

lelegard:
Note that the magic of operators allows to replace this:
~~~
    for (auto it = cat.descs.begin(); it != cat.descs.end(); ++it) {
        auto ca = CADescriptor(_duck, *it);
~~~
by this:
~~~
    for (const auto& desc : cat.descs) {
        auto ca = CADescriptor(_duck, desc);
~~~
That was the idea behind the addition of iterators. Explicit use of iterators are hardly better than the existing index `[]` operator.

mallikakotian7:
Hi @stazio any plans to merge the feature  to the release?

stazio:
Hey there,
So I noticed that @lelegard actually  started working on his own TR 101-290 analyzer in [master](https://github.com/tsduck/tsduck/blob/972c182182392ce5442bb8c93d0e5650e35ef979/src/libtsduck/dtv/tr101290/tstr101290Analyzer.cpp). I am excited to give it a try, if I ever get the time to come back to this.

Ultimately I stopped down on this because I wasn't really happy with the way it was implemented, and because I ran out of time and got pulled into other things. I do apologize for ghosting.

@lelegard do you have any problem if I close this out?

mallikakotian7:
hi @lelegard are you going to add tr-101-290 to analyze plugin too? Thanks!

stazio:
You should be able to take my src/tsplugins/tsplugin_tr101_290.cpp file and modify it to work with Thierry's analyzer.

If I ever get back to this (which I want too), I will look into making it work with a tsp plugin.

commdk:
> //! The "TS_sync_loss" and "Sync_byte_error" indicators are inaccessible in most cases. It depends
> //! on the input plugin. Many plugins, such as "dvb" or all IP-based plugins ("ip", "srt", "rist", etc.)
> //! use the 0x47 sync byte to resynchronize in the data stream or to locate TS packets in datagrams.
> //! Therefore, with these input plugins, corrupted sync bytes are filtered upstream and never reach
> //! the analyzer classes.

@lelegard Quick question what input plugins would show TS_sync_loss and Sync_byte_error?

PR #1592, opened 2025-03-18
use standard way to widen strings in macros

robUx4:
MSVC may be OK with the current code, but Clang is not. There is a macro to help with that.


lelegard:

PR #1593, opened 2025-03-18
tsPlatform: don't define TS_GCC for clang-cl

robUx4:
This is the Clang version with MSVC compatibility. It defines `_MSC_VER` just like MSVC and avoids gcc specific features in Clang.


lelegard:

PR #1594, opened 2025-03-18
fix __in usage if it's not defined

robUx4:
It's not defined by default when including Windows headers in mingw-w64.


lelegard:

PR #1595, opened 2025-03-18
SharedLibrary: silence error in function pointer case in LLVM

robUx4:
A recent LLVM complains when casting function pointers.

> base/system/tsSharedLibrary.cpp:122:18: error: assigning to 'void *' from 'FARPROC' (aka 'long long (*)()') converts between void pointer and function pointer

`static_cast` doesn't work so we use a C-style cast. The `getSymbol()` method is not used anyway.

### DIFF
--- a/src/libtscore/system/tsSharedLibrary.cpp
+++ b/src/libtscore/system/tsSharedLibrary.cpp
@@ -119,7 +119,7 @@ void* ts::SharedLibrary::getSymbol(const std::string& name) const
 #if defined(TSDUCK_STATIC)
         // Nothing to do, load() previously failed.
 #elif defined(TS_WINDOWS)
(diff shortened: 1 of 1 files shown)


robUx4:
The equivalent warning with clang-cl is:

```
[build] S:/Programs/tsduck/src/libtscore/system/tsSharedLibrary.cpp:122:18: warning: implicit conversion between pointer-to-function and pointer-to-object is a Microsoft extension [-Wmicrosoft-cast]
[build]   122 |         result = ::GetProcAddress(_module, name.c_str());
[build]       |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build] 1 warning generated.
```

lelegard:

PR #1596, opened 2025-03-18
use lowercase Windows headers

robUx4:
This is necessary to cross-compile on Linux with `mingw-w64`. `mingw-w64` use only lowercase for Windows headers. Using the Microsoft documented casing fails to build on file systems that are case sensitive. Using lowercase has the most compatibility among Windows build systems.

### DIFF
--- a/src/libtscore/crypto/tsSystemRandomGenerator.h
+++ b/src/libtscore/crypto/tsSystemRandomGenerator.h
@@ -17,7 +17,7 @@
 
 #if defined(TS_WINDOWS)
     #include "tsBeforeStandardHeaders.h"
-    #include <Wincrypt.h>
+    #include <wincrypt.h>
     #include "tsAfterStandardHeaders.h"
 #endif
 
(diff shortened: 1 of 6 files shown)


lelegard:
> This is necessary to cross-compile on Linux with mingw-w64. mingw-w64 use only lowercase for Windows headers. Using the Microsoft documented casing fails to build on file systems that are case sensitive. 

I would say it is a bug of the cross-compilation environment to provide Windows header files the name of which do not follow the Microsoft documentation.

Anyway, I will integrate this as is.

lelegard:

robUx4:
> I would say it is a bug of the cross-compilation environment to provide Windows header files the name of which do not follow the Microsoft documentation.

The documentation may not have been consistent throughout the years. wincrypt.h in the Windows SDK is in lower case. objidl.h is not, but can be found in lowercase in the [Microsoft documentation](https://learn.microsoft.com/fr-fr/windows/win32/api/objidl/nf-objidl-istream-seek).

PR #1600, opened 2025-03-18
tsCommandLine: avoid calling private method in default destructor

robUx4:
The method may be inlined in parent classes, in external code. That results in needing to export the private `PredefinedCommands` destructor so it can be called by the external code, even if it's not supposed to know about it.

Moving the destructor to the C++ file ensures the default destructor call is not inline and thus accessing the private `PredefinedCommands` destructor which is not DLL exported.

### DIFF
--- a/src/libtscore/app/tsCommandLine.cpp
+++ b/src/libtscore/app/tsCommandLine.cpp
@@ -19,6 +19,9 @@ ts::CommandLine::CommandLine(Report& report) :
 {
 }
 
+ts::CommandLine::~CommandLine()
+{
+}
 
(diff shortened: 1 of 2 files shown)


lelegard:
I have implemented a slightly different solution. Instead of implementing an explicit non-inline destructor, I added a `TSCOREDLL` attribute on the private inner class which is implicitly referenced in the destructor. This preserves the benefit of inlining while (hopefully) fixing the export issue.

Please check if it works for you with clang on Windows. If it does, please close this PR.

robUx4:
Yes it seems to be working fine.

PR #1601, opened 2025-03-18
DRAFT: add a CMake based makefile to build

robUx4:
This is an initial draft of something working for me for Windows and Linux targets.

It doesn't include the whole binaries and doesn't handle tests for now.

[file(GLOB)](https://cmake.org/cmake/help/latest/command/file.html#glob) could probably be used rather than hardcoding every file.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -52,88 +52,88 @@ on:
 
 jobs:
 
-  build_unix:
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-24.04, macos-latest]
-        compiler: [gcc, clang]
-        # Currently disable C++23 compilation, general support in compilers is too poor.
-        # std: [20, 23]
-        std: [20]
-        include:
-        - os: ubuntu-24.04
-          make: make DEBUG=${{ inputs.DEBUG }} VERBOSE=${{ inputs.VERBOSE }}
-        - os: ubuntu-24.04
-          compiler: clang
-          llvm: true
-        - os: macos-latest
-          # It has been noted on GitHub macOS runners that the host name can change
-          # between steps or even the middle of a step. This breaks the naming scheme
-          # of the build directory for binaries. As a consequence, we force BINDIR=bin.
-          bindir: bin
-          make: gmake DEBUG=${{ inputs.DEBUG }} VERBOSE=${{ inputs.VERBOSE }}
-        exclude:
-        - os: macos-latest
-          compiler: gcc
-          # Don't use gcc on macOS.
-    name: Build on ${{ matrix.os }} with ${{ matrix.compiler }}, C++${{ matrix.std }}
-    runs-on: ${{ matrix.os }}
-    env:
-      LLVM: ${{ matrix.llvm }}
-      BINDIR: ${{ matrix.bindir }}
-      CXXFLAGS_STANDARD: -std=c++${{ matrix.std }}
-      ORIGIN_REPO: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
-      REFERENCE_REPO: tsduck/tsduck
-      BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    steps:
-    - uses: actions/checkout@master
-    - name: Install dependencies
-      run: |
-        scripts/install-prerequisites.sh
-        ${{ matrix.compiler }} --version
-        scripts/java-config.sh
-    - name: Build TSDuck
-      run: ${{ matrix.make }} -j5
-    - name: Check built version
-      run: ${{ matrix.make }} show-version
-    - name: Run unitary tests
-      run: ${{ matrix.make }} test
-    - name: Download test suite
-      run: |
-        # Try to download '<repo>-test' from the pull request owner, if there is one.
-        if curl -fsL "https://github.com/${ORIGIN_REPO}-test/tarball/${BRANCH_NAME}" -o test.tgz; then
-            echo "Downloaded test suite from ${BRANCH_NAME} branch of $ORIGIN_REPO"
-        elif curl -fsL "https://github.com/${ORIGIN_REPO}-test/tarball/master" -o test.tgz; then
-            echo "Downloaded test suite from $ORIGIN_REPO"
-        else
-            curl -fsL "https://github.com/${REFERENCE_REPO}-test/tarball/master" -o test.tgz
-            echo "Downloaded test suite from $REFERENCE_REPO"
-        fi
-        mkdir -p ../tsduck-test
-        tar -C ../tsduck-test -x -z -f test.tgz --strip 1
-    - name: Run test suite
-      run: |
-        ${{ matrix.make }} test-suite && status=$? || status=$?
-        cd ../tsduck-test
-        for f in $(find tmp -name '*.diff'); do
-            echo "==== $f";
-            cat "$f"
-        done
-        exit $status
-    - name: Build sample applications
-      run: |
-        # Fake installation in a temporary directory.
-        export PREFIX=$(pwd)/bin/systest
-        ${{ matrix.make }} install SYSPREFIX=$PREFIX
-        # Build the test applications using that installation.
-        export PATH=$PREFIX/bin:$PATH:
-        export LD_LIBRARY_PATH=$PREFIX/lib:$LD_LIBRARY_PATH:
-        export LD_LIBRARY_PATH2=$LD_LIBRARY_PATH
-        export PYTHONPATH=$PREFIX/share/tsduck/python:$PYTHONPATH:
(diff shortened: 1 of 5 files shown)


lelegard:
Hi @robUx4 

I have merged 8 out of 10 PR's that you separately posted and which are present in this one. To avoid one additional merge commit per PR, I rebased them first. This has the unfortunate side effect that GitHub is unable to flag those PR as "merged" and I was obliged to manually close them. However, don't worry, your commits are here (the contributors sections on the code page already accounts 9 commits from you).

For the two additional PR's (inline methods referencing private inner classes), I have used another solution and didn't merge your commits. Please check if it works for you.

Concerning this draft PR about CMake, the hardly maintainable list of source files is exactly what I don't want to see in TSDuck.

> file(GLOB) could probably be used rather than hardcoding every file.

That is precisely what the "Modern CMake" zealots considered as a no-go. They explained that this "does not work" but their explanations were not very convincing for a CMake newbie like me. So, I am a bit perplex.

I see two difficulties with CMake. I don't know if they can be solved.

1. There are a lot of complex situations, depending on the OS, Linux, macOS, and the four BSD. Each OS is a specific case in terms of distro setup, compiler, header files, libraries, Homebrew or not. Some configs are very specific (Dektec, Vatek, SRT, RIST). Making it work on all OS will take time and, honestly, I don't feel like investing in this. I have 15 virtual machines on an Intel laptop and 5 virtual machines on an Arm Mac to test all of them. The current complexity is isolated in `make-config.sh` (with doc in `CONFIG.txt`).
2. With one global MSBuild solution, instead of one per project as generated by CMake, we can use Visual Studio as IDE on Windows (I use Qt Creator on Mac and Linux). CMake is fine when you have only one executable in your project. Here, TSDuck has more than 120 projects in the solution (libs, plugins, tools). Having 120 separate solutions, as probably generated by CMake, is fine for build automation, not for use with an IDE.

So, if we keep the GNU Make and MSBuild systems, using CMake is only there to use clang on Windows and I don't see any benefit. This is just added complexity, and just for a "non-natural" build combination which is not used outside your project.

robUx4:
> That is precisely what the "Modern CMake" zealots considered as a no-go. They explained that this "does not work" but their explanations were not very convincing for a CMake newbie like me. So, I am a bit perplex.

Yes this is frowned upon. For one thing if you add or remove a file, CMake (and in the underlying build system) won't notice and will not update the build accordingly. This is not a big deal in the CI, it becomes a problem if you switch between branches locally to develop. But running CMake is quite fast so it's not a major problem.

> There are a lot of complex situations, depending on the OS, Linux, macOS, and the four BSD.

The complexity about the varying toolchains is usually hidden in a [CMAKE_TOOLCHAIN_FILE](https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html). In VLC we build for a [lot of different platforms](https://code.videolan.org/fuzun/vlc/-/pipelines/576105) and run CMake and meson on all of them for our "contribs". Once you have a setup working for a platform, the rest is straightforward. For Android the NDK provides the toolchain file so there is nothing special to do (apart from picking the Android version you target). Android is one of the targets we will probably use tsduck with.

For the dependencies, there are standard ways of finding them. This is even easier when they were built with CMake (like SRT). Alternatively it's possible to plug a package manager into it like [vcpkg](https://github.com/microsoft/vcpkg/) which uses CMake too. You can use the existing dependencies or add your own and build exactly with what you want, from source.

CMake can also download tarballs to build "contribs", so it could download SRT, RIST, etc if you don't have them and build with them.

> Having 120 separate solutions, as probably generated by CMake, is fine for build automation, not for use with an IDE.

I think it generates one .`sln` file per "project()". In this case there's only one project. But modern Visual Studio can open CMake projects directly and let you pick the configuration you want and switch between them. I personally use VSCode which also has good CMake integration. I just open a folder and it's ready to build, which the possibility of picking the build configuration (Debug, Release, etc) and the compilers (all the ones it detects on the system).

robUx4:
> Do you think that it would be feasible to have a generic CMake config which would replace the current GNU Make + MSBuild system for TSDuck?

From what I've seen both from the makefiles and the VS project that is doable. I haven't looked at your PowerShell scripts to add the third party code. Even these could probably be merged in the CMake part in a more portable way.

> And, BTW, I am still curious about the reasons for a VLC / TSDuck integration and how you plan to implement that. Do you plan to keep the standard TSDuck modularity (libtscore + libtsduck + configuration files + plugin shared objects + executable tools) or do you plan to build a big library of what you need only?

The VLC integration would be through another library that uses tsduck, VLC itself would not know about tsduck (apart from the linking parts it needs to build). All the contribs we build are built as static library so we don't even care too much about the DLL aspect/fixes. Also we have patches for a lot of contribs although we usually try to "upstream" as much as possible so we don't have to maintain local patches and ensure it's correct according to the upstream project. One of the thing we often have to patch is the build system that can't handle all the various targets we use. So it would not be uncommon to have our own CMake file aside from your code. But if it can be "upstreamed" that would be a lot easier for everyone.

robUx4:
I removed the parts to build with mingw-w64 so there's only the CMake project and test of that project in the CI (winget seems to fail on some of them, not sure why).

lelegard:
> (winget seems to fail on some of them, not sure why).

This is a recurrent issue on GitHub CI runners. They do not integrated winget by default. The installation of winget sometimes fails for some unknown reasons. I try to make the script `install-winget.ps1` more robust (retries, etc) but it is hard to test it in the random failure cases.

robUx4:
I added a first `file(GLOB)` to get the list of plugins. That went well.

robUx4:
I did the rest of the file globbing to be closer to the other build systems and be more flexible. There's still a lot missing but at least it helps checking it build with clang-cl, in addition of the existing CI targets.

lelegard:
> Also we have patches for a lot of contribs although we usually try to "upstream" as much as possible so we don't have to maintain local patches and ensure it's correct according to the upstream project.

I agree with that. About the C++ code, I am ready to merge any patch which improves the code. This is good for the project. Feel free to submit PR's when necessary.

About the build system, I am open to consider a migration from GNU Make / MSBuild to CMake if we can get the same features, flexibility, and list of targets as the current TSDuck build system. If it covers only a subset of the needs, then I prefer that you keep it and maintain it in your project.

The problem with separate build systems is that I cannot reproduce your builds, cannot run non-regression tests in your configuration and cannot guarantee that some future code change will not break your build.

robUx4:
> About the build system, I am open to consider a migration from GNU Make / MSBuild to CMake if we can get the same features, flexibility, and list of targets as the current TSDuck build system. If it covers only a subset of the needs, then I prefer that you keep it and maintain it in your project.

I understand and I will do that. The problem is that we only need a subset of what tsduck does. So putting extra work to support every single thing that tsduck can do is not on the map for now.

> The problem with separate build systems is that I cannot reproduce your builds, cannot run non-regression tests in your configuration and cannot guarantee that some future code change will not break your build.

As said earlier, we could have the CMake build system on the side, as not officially supported, but nightly (or weekly) and updated/improved gradually. From what I've seen CMake should be capable of handling of these your 2 build systems do, including Python and Java integration.

I will keep updating this branch when I can, it should at least build and pass the tests.

lelegard:
@robUx4,

> All the contribs we build are built as static library so we don't even care too much about the DLL aspect/fixes

There is one pitfall you should be aware of when putting TSDuck code into static libraries.

Given the constraints that 1) I don't have much time and 2) there is a lot of code into the project with regular additions, I pay attention to remove all redundancies because they need extra work and are error prone.

No hard-coded lists of files is one example: Adding a new source file shall be self-sufficient. Updating a file list is redundant extra work which will be easily overlooked.

Another example is self-registration of C++ classes. New plugins, tables, descriptors, extension features, are implemented as separate C++ classes. Adding the class in the code base shall be sufficient. No redundant list of initialization calls, tables of pointers, or external init sequence shall be necessary. We achieve this using self-registration macros ([explained here](https://tsduck.io/download/docs/tsduck-dev.html#selfreg)). As soon as a new C++ class for a descriptor is added, the implicit initialization code of the compilation unit automatically registers it into the signalization repository. This is very convenient, no risk to forget some external initialization requirement.

The consequence of this system is that new features are _pushed_ into the code base. They are not _pulled_ by the application.

In the standard TSDuck build, all classes are automatically included in the relevant shared libraries, depending on the location of the source file. When the OS loader activates the shared library, all modules are initialized and all features are self-registered.

However, if all modules are placed in a static library, the linker will only include the objects which are explicitly pulled, using symbol references. Concerning the signalization, the application may explicitly reference a few common stuff such as a PAT, PMT or CA descriptor. However, the vast majority of the signalization will not be pulled. If your application analyzes the signalization of a stream, most of if won't be handled properly.

In the makefile of libtsduck, when building the static library, we build intermediate big object files gathering all tables, all descriptors, etc. So, from the application, as long as you reference one table for instance, you pull the big object which contains all of them. Using static libraries in TSDuck is not very common. We use for test only and because a few people wanted it to build on small embedded systems.

If you create your own build system based on static libraries only, you should keep this in mind.

PR #1608, opened 2025-04-02
tsEditLine: fix unused warning with NOEDITLINE

robUx4:
The unused warning will turn into a build error with -Werror

### DIFF
--- a/src/libtscore/app/tsEditLine.cpp
+++ b/src/libtscore/app/tsEditLine.cpp
@@ -40,8 +40,10 @@ ts::EditLine::EditLine(const UString& prompt, const UString& next_prompt, const 
     _is_a_tty(StdInIsTerminal()),
     _prompt(prompt),
     _next_prompt(next_prompt),
-    _history_file(history_file),
-    _history_size(history_size)
+    _history_file(history_file)
+#if !defined(TS_NO_EDITLINE)
+    ,_history_size(history_size)
+#endif
 {
 #if !defined(TS_NO_EDITLINE)
     if (_is_a_tty) {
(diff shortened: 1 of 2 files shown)


robUx4:
No sure why it's building with `Installing Command Line Tools for Xcode-15.3` ... when it should be building with `Xcode-16.2` as in https://github.com/tsduck/tsduck/actions/runs/14158591921/job/39661045031

lelegard:
I have fixed it using `[[maybe_unused]]` attributes. I will close the PR.

> No sure why it's building with Installing Command Line Tools for Xcode-15.3 ... when it should be building with Xcode-16.2

Not sure either. The `macos-latest` GitHub runners use macOS 14.7.4 while the latest is 15.3.2. That's why we must force an update of the command line tools to boost clang to version 16. However, this is not completely easy. The script `install-xcode-command-line-tools.sh` forces the update and you can see that the method is neither obvious nor fully reliable.

It seems that all CI runs on macOS today fail with the same error.

I see this on my mac (macOS 15.3.2):
~~~
$ softwareupdate --list
Software Update Tool

Finding available software
Software Update found the following new or updated software:
* Label: Command Line Tools for Xcode-16.3
        Title: Command Line Tools for Xcode, Version: 16.3, Size: 861139KiB, Recommended: YES,
* Label: macOS Sequoia 15.4-24E248
        Title: macOS Sequoia 15.4, Version: 15.4, Size: 6021527KiB, Recommended: YES, Action: restart,
~~~

I wonder how it looks like on a macOS 14. Maybe (just maybe) the new Command Line Tools for Xcode-16.3 is not available for macOS 14. I will try to force the CI jobs on `macos-15` runners.

lelegard:
Forcing mac CI on `macos-15` runners fix the issue with "Clang 16 is already installed".

PR #1609, opened 2025-04-02
tsDekTec: disable some warnings when including DTAPI.h

robUx4:
At least with GCC 14.2.0.

This will fail to compile with `-Werror`.

### DIFF
--- a/src/libtsduck/dektec/private/tsDektec.h
+++ b/src/libtsduck/dektec/private/tsDektec.h
@@ -47,6 +47,12 @@ TS_MSC_NOWARNING(4626)
 TS_MSC_NOWARNING(5026)
 TS_MSC_NOWARNING(5027)
 TS_MSC_NOWARNING(5204)
+TS_GCC_NOWARNING(non-virtual-dtor)
+TS_GCC_NOWARNING(overloaded-virtual)
+TS_GCC_NOWARNING(zero-as-null-pointer-constant)
+TS_LLVM_NOWARNING(non-virtual-dtor)
+TS_LLVM_NOWARNING(overloaded-virtual)
+TS_LLVM_NOWARNING(zero-as-null-pointer-constant)
 #define _NO_USING_NAMESPACE_DTAPI
 #include "DTAPI.h"
 TS_POP_WARNING()


lelegard:
Hi @robUx4 

I don't see these warnings on Ubuntu 24.10 with GCC 14.2.0. In which context do you see the warnings? In your port to Windows with GCC?

I will merge this one because it does not hurt and it's fine if it helps you, but I prefer to understand first.

I don't know what would be the result of building on Windows with gcc or clang and Dektec support. The DTAPI is provided in binary form only. In their WinSDK, the DTAPI is provided as a static library for MSVC. I assume that this code can be called from gcc/clang code but I have no idea of the compatibility of the header file DTAPI.h, beyond the warnings you got. As a precaution, I would recommend to build with `NODTAPI=1` when building for Windows using gcc or clang. It may work but it won't be supported by Dektec. Do you need to use Dektec devices?

robUx4:
> I don't see these warnings on Ubuntu 24.10 with GCC 14.2.0. In which context do you see the warnings? In your port to Windows with GCC?

This is on Debian sid with gcc 14.2.0, using the CMake build. I'm not enabling all the same warning as you do in your build system but if anything I'm using less strict flags, so not sure why I have the issue and you don't.

robUx4:
> As a precaution, I would recommend to build with NODTAPI=1 when building for Windows using gcc or clang. It may work but it won't be supported by Dektec. Do you need to use Dektec devices?

Yes I noticed the CMake unpacking of the ZIP is not sufficient on Windows so I turned it off by default with CMake (and issue an error if it's enabled). It's possible to call the .ps1 script to install it via CMake. But that can only be used on Windows or with Linux with Wine. And even then the provided C++ MSVC libraries will likely not link with mingw-w64 toolchains. So It will probably never be possible to cross-compile with that feature.

lelegard:
> This is on Debian sid with gcc 14.2.0, using the CMake build. I'm not enabling all the same warning as you do in your build system but if anything I'm using less strict flags, so not sure why I have the issue and you don't.

Weird indeed. I merge it anyway.

> Yes I noticed the CMake unpacking of the ZIP is not sufficient on Windows so I turned it off by default with CMake (and issue an error if it's enabled).

From Dektec, the LinuxSDK and WinSDK are quite different. In the make system, we download and unzip the LinuxSDK. It works only on 1) Linux, 2) Intel target, 3) using glibc (i.e. not Alpine Linux since it uses muslc). In all other configurations, Dektec support is disabled.

The WinSDK is an archive which contains a binary installer for Windows. So, installing it is part of `install-prerequisites.ps1` instead of the build.

PR #1611, opened 2025-04-03
Fix misc. warnings found with GCC + mingw-w64

robUx4:
I can split into multiple PRs if it's easier to merge.


lelegard:
Well done :) 

Most of this code is rather old, from a time where `nullptr` did not exist and `0` was the recommended null pointer literal in C++. After C++11, the code went through a lot of refactoring but the Windows-specific code was only compiled with MSVC, which does not report these as warning.

Thanks.

PR #1614, opened 2025-04-08
tsCOM: remove unused comsuppw library

robUx4:
It's not available in mingw-w64 and none of the helpers provided are used.


lelegard:
Hi @robUx4 

Well spotted. I think I know where it comes from. At the beginning, TSDuck used 8-bit strings `std::string` and `ConvertBSTRToString` (from commsuppw) was used to convert from COM strings. When TSDuck moved to 16-bit strings `UString`, the conversion became straightforward and `ConvertBSTRToString` was no longer necessary but the reference to commsuppw remained.

PR #1623, opened 2025-04-28
tsThread: don't call SetThreadDescription() is compiled for older Windows

robUx4:
The documentation [^1] says the API is available since Windows 10 1607, aka Redstone 1. To support it regardless of the compilation flags it should be loaded dynamically.

[^1]: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription#requirements


lelegard:
Hi @robUx4 

I implemented your suggestion to dynamically load SetThreadDescription to make the binary independent of the version of Windows on which it is run.

If it works for you, I let you close this one.

robUx4:
Yes it's cleaner when dynamically loaded.

PR #1640, opened 2025-06-05
tsPlatform.h: fix powerpc macros

barracuda156:

Trivial change, fix powerpc detection. No impact on supported platforms.

### DIFF
--- a/src/libtscore/cpp/tsPlatform.h
+++ b/src/libtscore/cpp/tsPlatform.h
@@ -378,11 +378,11 @@
     #if !defined(TS_SPARC)
         #define TS_SPARC 1
     #endif
-#elif defined(__powerpc64__)
+#elif defined(__powerpc64__) || defined(__ppc64__)
     #if !defined(TS_POWERPC64)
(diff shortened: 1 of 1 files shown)


barracuda156:
> Thanks. I doubt that anyone will use TSDuck on AIX. Does anybody still use AIX?

AIX can run on the latest POWER cpus, I believe. And at least corporate should have long-term contracts with IBM, so some folks may be “locked into” using AIX even if they rather prefer something else. If I could afford a Power9 machine, I would try AIX for sure LOL

barracuda156:
@lelegard That’t really impressive!

I still hope for OpenPower to provide something cheaper than Raptor Talos. (I was stupid enough not to buy it before 2020, but who knew.)

But I am kinda fine with RISC-V. Once a usable laptop on it reaches the market, I will consider it. Sadly, they make it LE, though the arch supports BE, AFAIK.

P. S. On a side note, what’s wrong with OpenBSD? It is well alive. Better supported on some archs than FreeBSD. And NetBSD also alive.

PR #1650, opened 2025-07-15
Add Unicable support to tsp dvb (input) plugin.

massweet:
#### Related issue (if any):
#1101

#### Affected components:
dvb (input) plugin

Add support for Unicable I and II to linux's version of the dvb (input) plugin (EN50494 and EN50607 respectively).

Modify some of the documentation about "satellite-number" to clarify which hardware has what ranges. In particular, as we are adding non-DiSEqC equipment to what is supported clarify the "non-DiSEqC" switches with a satellite number range of 0..1 to being DiSEqC tone-burst switches.

Also clarify that what is currently referred to as a "DiSEqC switch" is usually a "DisEqC 1.0" switch.

Test setup:

A server running Ubuntu 24.04, with a TBS dual tuner satellite card connected via adapter 1 to a Johansson 9775 Unicable multiswitch that in turn is connected to four satellites via wideband LNBs.

Testing:

Both Unicable I and Unicable II code paths successfully tested while blind-scanning the entire frequency/polarity range of all four satellites attached to the Unicable switch.

### DIFF
--- a/doc/user/plugins/dvb-input.adoc
+++ b/doc/user/plugins/dvb-input.adoc
@@ -533,7 +533,13 @@ Used for satellite tuners only.
 
 [.optdoc]
 Satellite/dish number.
-Must be 0 to 3 with DiSEqC switches and 0 to 1 for non-DiSEqC switches.
+Valid range is:
+
+* DiSEqC tone burst switch: 0 to 1;
+* DiSEqC 1.0 switch: 0 to 3;
+* Unicable I switch: 0 to 1;
+* Unicable II switch: 0 to 63;
+
 The default is zero.
 
 [.opt]
--- a/src/libtsduck/dtv/broadcast/linux/tsTunerDevice.cpp
+++ b/src/libtsduck/dtv/broadcast/linux/tsTunerDevice.cpp
@@ -1057,9 +1057,9 @@ bool ts::TunerDevice::dishControl(const ModulationArgs& params, const LNB::Trans
 
     // Send tone burst: A for satellite 0, B for satellite 1.
     // Notes:
-    //   1) DiSEqC switches may address up to 4 dishes (satellite number 0 to 3)
-    //      while non-DiSEqC switches can address only 2 (satellite number 0 to 1).
-    //      This is why the DiSEqC command has space for 2 bits (4 states) while
+    //   1) DiSEqC 1.0 switches may address up to 4 dishes (satellite number 0 to 3)
+    //      while tone-burst switches can address only 2 (satellite number 0 to 1).
+    //      This is why the DiSEqC 1.0 command has space for 2 bits (4 states) while
     //      the "send tone burst" command is binary (A or B).
     //   2) The Linux DVB API is not specific about FE_DISEQC_SEND_BURST. Reading
     //      szap or szap-s2 source code, the code would be (satellite_number & 0x04) ? SEC_MINI_B : SEC_MINI_A.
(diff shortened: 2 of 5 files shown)


lelegard:
Hi @massweet 

Thank you for your work on this.

I have a few questions or remarks. If I understand correctly, the intermediate frequency on which the tuner is programmed is set in the `--unicable` parameter, right? In that case, the `--frequency` parameter becomes unused. 

This also means that the user has to exactly know the intermediate frequency for each documented satellite frequency, on each model of Unicable switch. This is exactly what I wanted to avoid with LNB's. The user only knows the publicly documented satellite frequencies and the model of his LNB (or Unicable switch). Then, TSDuck should hide the guts.

So, to remain in the TSDuck mindset, we should:

- Extend the concept of LNB to Unicable switch features. This probably means extend the syntax of the predefined LNB XML files and the corresponding logic in the code.
- Add the configuration of some known Unicable switches in the LNB XML configuration file.
- Use the name of the Unicable switch in the `--lnb` option (which becomes slightly misnamed but the plugin `dvb` became misnamed as well when I extended it to ATSC and ISDB).
- Extend the C++ class `LNB` to support Unicable switches and compute intermediate frequencies from satellite frequencies using the configuration in the XML file. Also expose the Unicable nature of an "LNB" to send the right DiSEqC commands.
- Remove the `--unicable` option which becomes unnecessary because Unicable is just a specific case of frequency translation system (generalization of LNB), driven by `--frequency`, `--lnb` and `--satellite-number`.

Does it make sense?

It may seem complicated but necessary. TSDuck is complicated: 500,000 lines of codes, managing large combinations of standards, configurations, exceptions, etc. The robustness and scalability of the code are keys. If we add specific concepts all the time, we will end up with unmaintainable spaghetti-plate code. This is why I have always made all efforts to generalize existing concepts when new features are needed and introduce new concepts when necessary only.

And here, Unicable switches are just "frequency translation systems", like LNB's. So, let's extend the concept. Note that, initially, the concept of LNB was simple. When I started to support ISDB-S with the help of Japanese users, I realized that many new parameters and polarizations should be added. Complicating the parameters was a bad idea because it exposed the internals of the LNB models to the users. So, I rewrote the LNB system with XML configuration files. Now, let's extend it once more.

lelegard:
If I understand correctly, there is only one frequency (the user band) between a receiver and the switch, right? And that means that a receiver cannot have multiple tuners to demodulate several transponders (provided that they are on the same quarter) because it can receiver only one frequency.

So, that is why you need to specify the user frequency, because each receiver has a statically assigned user frequency, right? Several receivers in the same building, receiving the same satellite frequency, will receive that same stream on distinct frequencies, its statically user frequency. It this correct?

Another question: In your code, you manually compute the intermediate frequency, between the switch and the LNB. Your code uses the parameters of the "European Universal Ku (extended)" LNB, as defined in the LNB description file, ignoring the value of `--lnb`. Does this mean that Unicable must have that type of LNB and will not work with any other?

Also consider that TSDuck uses its own string class everywhere, `UString`, a unicode string, with features to split strings and convert to integers.

lelegard:
> Just in case you think it is not used in the UK, Unicable is the technology underlying dCSS/dSCR. Anyone with a SKY Q box set to use SCR mode should be able to use this (and the user-band to frequency mapping you can find with a bit of googling) to use a capture-PC alongside their SKY infrastructure.

My experience of UK TV is limited to terrestrial, when I worked in London in 1998 for the launch of the UK DTTV. It must have changed quite a bit since then.

I had never heard of Unicable before your PR (including your colleague's work).

massweet:
We should also probably turn the user-band off at the end of using the plugin. See #1656 for details.

massweet:
I think I should rework this MR after the DiSEqC 1.1 one goes in - There will be conflicts between them as I have changed some of the same comments.

massweet:
...where (if anywhere) can I see your changes?

massweet:
Ah. OK. just a bit of code inspection suggests a need to set `_voltage_on = true;` so it gets turned off again on exit.

lelegard:
> code inspection suggests a need to set _voltage_on = true

Thanks, fixed.

massweet:
...still trying to get back to this.

Issue #1651, opened 2025-07-16
[BUG] Build failure: "AttributeError: module 'xml.etree.ElementTree' has no attribute 'indent'"


murdoa:
#### Bug description:

Failure when building tsduck due to missing feature (xml.etree.ElementTree.indent) within xml Python library. Occurs when running "build-tables-model.py" as part of the build process.

xml.etree.ElementTree.indent was added in Python 3.9. I have not noticed a required minimum Python version within the docs. This build succeeds when omitting the pretty printing step.

~~~
    build-tables-model.py
    49: # Pretty print the XML file.
    50: ET.indent(root)
~~~

[Link to build-tables-model.py:50](https://github.com/tsduck/tsduck/blob/93190529a7a70d5b7b12706d85f882fe82bd7b2e/scripts/build-tables-model.py#L49C1-L50C20)

#### How to reproduce:

I have encountered this bug on two builds for separate systems both with Python versions <3.9.

#### Errors and logs:

~~~
Traceback (most recent call last):
  File "tsduck/scripts/build-tables-model.py", line 50, in <module>
    ET.indent(root)
AttributeError: module 'xml.etree.ElementTree' has no attribute 'indent'
~~~

#### Environment:
 - OS:  Ubuntu
 - OS version:  24.04
 - TSDuck full version: v3.41-4299
 - Installation type: Source

PR #1652, opened 2025-07-16
Fix build failures with Python<3.9 due to (xml.etree.ElementTree.indent)

murdoa:
#### Related issue (if any):
#1651 

#### Affected components:
Python build script: [build-tables-model.py](https://github.com/tsduck/tsduck/blob/master/scripts/build-tables-model.py)

When building with Python versions <3.9 a build error occurs as the xml library within Python's standard library is missing xml.etree.ElementTree.indent feature.

I've proposed a fix here which checks if this feature exists and will generate the pretty print tree or print a warning message stating "Not pretty printing XML (not supported by Python<3.9)".

### DIFF
--- a/scripts/build-tables-model.py
+++ b/scripts/build-tables-model.py
@@ -47,7 +47,10 @@ if __name__ == '__main__':
     for topic in root:
         topic[:] = sorted(topic, key=lambda e: e.tag.lower() if isinstance(e.tag, str) else '')
     # Pretty print the XML file.
-    ET.indent(root)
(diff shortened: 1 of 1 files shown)


lelegard:
Hi @murdoa,

Thanks a lot for finding and fixing this.

BTW, I had to fix the fix because you wrote `if hasattr(ET, 'ident')`, "ident" instead of "indent", and the pretty print was always skipped, even on the latest versions of Python.

Issue #1275, opened 2023-08-01
[Q] How to enter the sat position for DiSEqC 1.1?


andimik:
In the PDF, I read

> --satellite-number
value
Used for satellite tuners only.
Satellite/dish number. Must be 0 to 3 with DiSEqC switches and 0 to 1 for non-DiSEqC
switches. The default is zero.

But a friend of mine wants to use DiSEqC 1.1 as he has up to 16 different sat positions.

https://en.wikipedia.org/wiki/DiSEqC

lelegard:
Hi @andimik 

There is currently no support for more recent versions of DiSEqC. The current limit is 4 dishes. 

I keep this suggestion for improvement. However, I will not be able to test this kind of reception. So, the improvement will be implemented only if there is a commitment from a user to test it and provide feedback. 

The improvement will be implemented on Linux only. On Linux, the DiSEqC commands are explicitly built by the application (TSDuck) and updates are possible. On Windows, this support is built inside DirectShow and BDA, the Windows framework for broadcasting which is more or less abandoned by Microsoft and I do not expect any improvement.

andimik:
Thank you for the fast response.

I forgot to tell that the OS is Ubuntu.

> However, I will not be able to test this kind of reception.

Me too, I will only install `tsduck` on his laptop, but he needs to test it.

But we think we have a workaround: A DVB-S(2) receiver with DiSEqC 1.1 support and LNB loop-through might help here. So the receiver tunes to that satellite and I expect that there is no need to use any DiSEqC parameters under Ubuntu.

andimik:
https://www.digitalfernsehen.de/ratgeber/workshop-bis-zu-64-sat-positionen-mit-diseqc-1-1-empfangen-555892/

The author of this article (and also of many other sat books) is the person where I wanted to try that and it's his sat antenna.

I'm not a programmer, just a user, but maybe that helps: https://github.com/Barracuda09/SATPI/issues/150

Or https://www.i-have-a-dreambox.com/wbb5/forum/index.php?thread/106589-probleme-mit-diseqc/&postID=911621#post911621

massweet:
Quite how DiSEqC tone burst, 1.0 and 1.1 switches "stack" or "cascade" is left as a user/installer choice.

The DiSEqC specifications permit cascading, but do not give guidance as to "approved configurations", indeed with DiSEqC bus 2.x capable devices and receivers the receiver can discover what is present, and change the slave address of devices leading to very complex installations.

[...]

PR #1655, opened 2025-07-30
Add control of DiSEqC 1.1 (uncommited) switch

massweet:
#### Related issue (if any):
#1275

#### Affected components:
dvb (input)

Add sending a port-group 1 command, followed by a 100ms delay, and move the tone-burst command after the port-group 0 command.

This lines up with the DisEqC bus specification sections 5.3 and 5.4.

satellite-number is now in the range 0..63. The DiSEqC 1.1 switch moves on every four satellite-numbers, and a cascaded DisEqC 1.0 switch then moves on within those four satellite numbers.

If no suitable switch is present, then the command is ignored.

Testing:

The receiver was attached to a SPAUN SUR420 WSG DiSEqC 1.1 switch, set to mode 1;
1. port 1 had a voltage indicator attached;
2. port 2 had a DS-04C DiSEqC 1.0 switch, with voltage indicators on: lnb1; lnb2; and lnb3.
3. ports 3 and 4 were not connected.

Attempts were made to tune to:
* no defined satellite number (spaun port 1);
* satellite-number 0 (spaun port 1);
* satellite-number 1 (spaun port 1);
* satellite-number 4 (DS-04C lnb1);
* satellite-number 5 (DS-04C lnb2);
* satellite-number 6 (DS-04C lnb3);
* satellite-number 8 (no indicators lit).

The above were all as expected.

It was noticed that the lnb voltage is not cleared to 0V if tsp is no longer running.

### DIFF
--- a/src/libtsduck/dtv/broadcast/linux/tsTunerDevice.cpp
+++ b/src/libtsduck/dtv/broadcast/linux/tsTunerDevice.cpp
@@ -423,6 +423,9 @@ void ts::TunerDevice::hardClose(Report* report)
         _demux_fd = -1;
     }
     if (_frontend_fd >= 0) {
+        // Attempt to turn off the LNB power
+        ioctl_fe_set_voltage(_frontend_fd, SEC_VOLTAGE_OFF);
+
         ::close(_frontend_fd);
         _frontend_fd = -1;
     }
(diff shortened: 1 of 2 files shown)


massweet:
See #1656 for the LNB continuing to be powered after the dvb input plugin has been torn down.

Issue #1677, opened 2025-10-16
[BUG] Descrambler failed with AES-128-CBC stream file with non-fixed cw


wangpeng2020:
#### Bug description:

when descrambler the AES-128-CBC stream file with non-fixed cw, the descrambler plugin report error and exit

#### How to reproduce:

1.  creat a scramble ts file.  use the attach shell scramble_local_aes.sh.
3.  use tsp to descramble   
 `../release-x86_64/tsp --debug=1 -r -I file -r 10 output/scrambled_test_aescbc.ts -P regulate -P descrambler 2  --aes-cbc -O ip 10.9.100.100:12345`
#### Expected behavior:

descramble  correctly

#### Errors and logs:

~~~
* Debug: trying to load "/opt/tsduck/release-x86_64/tsplugin_descrambler.so"
* Debug: registering processor plugin "descrambler", status: ok
* Debug: descrambler: ====> tsp -P descrambler 2 --aes-cbc
* Debug: descrambler: switching scrambling type from DVB-CSA2 to AES-CBC with externally-defined IV (TSDuck-specific)
* Debug: tsp: buffer size: 89,240 TS packets, 16,777,120 bytes
* Debug: regulate: starting bitrate-based regulation
* regulate: minimum packet burst duration is 10000000 nanoseconds
* Debug: descrambler: ECM processing thread started
* Debug: file: detected TS file format TS
* Debug: file: initial buffer load: 44,620 packets, 8,388,560 bytes
* file: initial input bitrate is 38,014,878 b/s
* Debug: ip: create IPv4 socket, type 2, protocol 17
* Debug: ip: binding socket to 0.0.0.0
* Debug: ip: setting UDP socket default destination to 10.9.100.100:12345
* Debug: ip: setting socket IP_MULTICAST_LOOP to 1
* Debug: file: input thread started
* Debug: regulate: packet processing thread started
* regulate: regulated at bitrate 38,014,878 b/s
* Debug: descrambler: packet processing thread started
* Debug: regulate: new regulation, burst: 10,000,000 nanoseconds, 253 packets, measurement period: 1,000,000,000 nanoseconds
* Debug: ip: output thread started
* Debug: control commands: found 2 packet processor plugins
* Debug: descrambler: adding standards MPEG to none
* descrambler: found service id 0x0002 (2), PMT PID is 0x0400 (1024)
* Debug: descrambler: PMT: service 0x0002, 2 elementary streams
[...]

PR #1678, opened 2025-10-17
Feature add sm4 crypto from openssl for scramble/descramble plugin

wangpeng2020:
#### Related issue (if any):

applied the fix of the issue
[Fixed IV init issue in descrambler (issue](https://github.com/tsduck/tsduck/commit/c9a59afda9531bc9e5e21e9fa82a6caa6fa070ce) https://github.com/tsduck/tsduck/issues/1677[)](https://github.com/tsduck/tsduck/commit/c9a59afda9531bc9e5e21e9fa82a6caa6fa070ce)
#### Affected components:

scramble/descramble plugin
add other crypto sm4 which is very same with aes-128.  it only suport on liunx with openssl
use --sm4-ecb、--sm4-cbc when use scramble/descramble plugin
#### Other
test pass on openEuler 24.09 with openssl-libs-3.0.12

### DIFF
--- a/src/libtsduck/dtv/cas/tsTSScrambling.cpp
+++ b/src/libtsduck/dtv/cas/tsTSScrambling.cpp
@@ -33,6 +33,12 @@ ts::TSScrambling::TSScrambling(const TSScrambling& other) :
     setScramblingType(_scrambling_type);
     _dvbcsa[0].setEntropyMode(other._dvbcsa[0].entropyMode());
     _dvbcsa[1].setEntropyMode(other._dvbcsa[1].entropyMode());
+    _aescbc[0].setIV(other._aescbc[0].currentIV());
+    _aescbc[1].setIV(other._aescbc[1].currentIV());
+    _aesctr[0].setIV(other._aesctr[0].currentIV());
+    _aesctr[1].setIV(other._aesctr[1].currentIV());
+    _sm4cbc[0].setIV(other._sm4cbc[0].currentIV());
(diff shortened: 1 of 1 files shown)


lelegard:
Hi @wangpeng2020 

Thank you for contributing !

Could you please provide the Windows alternative? TSDuck is multiplatform and we make all possible effort to have the same level of feature on all operating systems.

TSDuck chose to use the "native library" for a given feature on each operating system. On all UNIX systems (Linux, macOS, BSD), OpenSSL is used for cryto. On Windows, we use the Microsoft BCrypt library, aka CNG (Crypto New Generation). If there is only one reason for that choice, try AES on Windows Arm64: BCrypt is 60 times faster than OpenSSL (because OpenSSL build system does not use Arm64 assembly code on Windows, while it does on UNIX).

For new source files, also use your full name in the copyright line. As [explained here](https://tsduck.io/docs/tsduck-dev.html#licensecopyright), "Open-source developers deserve credits for their work. The full real name of the author of a source file shall be mentioned in the copyright line of the header."

wangpeng2020:
thanks for your reply.  It's my first pr.   I  seen  tsduck using  Microsoft BCrypt library on windows, but seems it don't support sm4. 
so i just remove the code bout windows, sorry i didn't setup a windows compile env to test it.   Also consider the sm4  is very similar to the AES,  i thinks it' not a must requiement.  So i will close this pr.  And thanks a lot for your tsduck, it help s me a lot.

PR #1695, opened 2026-01-29
Adding flake.nix to repository.

jcekstrom:
#### Affected components:

Tests - Adds a NOEXTERNALTESTS flag to disable tests that require network and external file system access
 
This change adds support for Nix Flakes, and allows tsduck to be easily embedded in other flakes and installed into the NixOS, Nix-Darwin, etc.

### DIFF
--- a/pkg/nix/README.md
+++ b/pkg/nix/README.md
@@ -11,6 +11,19 @@ This directory contains the Nix packaging for TSDuck and its optional hardware d
 - **`python-tsduck`** / **`python-tsduck-min`** — Python bindings
 - **`java-tsduck`** / **`java-tsduck-min`** — Java bindings
 
+### Overlay
+
+The flake exports `overlays.default` which adds the following to nixpkgs:
+
+- **`pkgs.tsduck`** — Full TSDuck C++ package
+- **`pkgs.tsduck-min`** — Minimal TSDuck (no hardware SDKs)
+- **`pkgs.python3Packages.tsduck`** — Python bindings (full)
+- **`pkgs.python3Packages.tsduck-min`** — Python bindings (minimal)
+
+The Python packages are injected via `pythonPackagesExtensions`, making them
+available under all Python interpreters (e.g. `python311Packages`,
+`python312Packages`, `python313Packages`) and work with `python3.withPackages`.
+
 ## Building
 
 ### From GitHub (flake URL):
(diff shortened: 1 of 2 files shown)


jcekstrom:
> Another point on `flake.lock`. It contains precise references of dependent packages, including hash values. I assume that this is why it is named "lock".
> 
> This kind of version-specific, or version-locking, information has nothing to do in a source code repository for a project. The source code of a project is a rolling state which can always be built on various environments. It cannot be bound to a specific environment.

I agree.  The point of the flake.lock is that a build would be immutable, so when I tag this release, I know that build has these specific versions of dependencies, so it I install that flake from the tag on NixOS, I will _always_ get that exact build.  For a project like TSDuck I agree that it is less useful, and will remove the flake.lock.  Downstream flake consumers, say someone utilizing tsduck with the python scripting, or java, could lock the dependencies in their flake, that uses tsduck as an input.  

Already, flakes have a mechanism to override the "nixpkgs" version, which equates to a package dependency tree.

> There is a good example of do's and don't with `pkg/deb`. Using `make` in this directory builds a `.deb` package for TSDuck for the current operating system, whichever it is. If run on Debian 13, it builds a package for Debian 13, using the dependencies which are present on Debian 13. If run on Ubuntu 24.04, Ubuntu 25.10, or Linux Mint 22.3, it builds a package for that OS, with the corresponding dependencies.

> If this is possible (which I do not know), you should apply the same principle to Nix packaging. Put everything in `pkg/nix`, use templates for dependencies and use a Makefile or a script to build the exact dependencies and the package.
> 
> If this is not possible and incompatible with the spirit of Nix (because immutability is the core of it), then, sorry, but Nix won't make it into TSDuck source code repository. The source code repository can only provide tools to build and package the current version of TSDuck on any version of an operating system, not lock TSDuck to a specific version of it.

Agreed, and honestly a flake.lock gets ignored much of the time as when defining your inputs, most flakes do this:

```
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-<version>;
    tsduck = {
        url = "github:tsduck/tsduck/v3.43-4549";
        inputs.nixpkgs.follows = "nixpkgs";
    };
```
Which, according to my limited understanding, instructs tsduck to be build with the dependencies from <version> of the nixos system.

So I think we can just leave it out.  The flake.lock is most useful if we were creating and caching binary builds for people to use, then nix would check the inputs against the lock file, and determine whether it needed to build or could download a cached version for the specified dependencies.

Needless to say, my goal with this is to allow people to easily install and depend on tsduck in nix based systems.

I'll get you some updates that should address your concerns.

jcekstrom:
@lelegard I did have a question about your branches.  For the latest "stable" release, currently `v3.43-4549`, should I update the docs to point to the `develop` branch, as that seems to be tied to the latest stable release, or would you prefer I use explicit version, or just <latest version>

lelegard:
Hi @jcekstrom 

> There is an external repository, however often it is extremely behind the latest release, and does not have many of the flags enabled, like SRT and RIST support. I tried to get the package updated there, but after waiting for weeks for build to happen on the CI, I gave up.

OK, so if there is an external repository, let it reference immutable versions, and this is a good reason to have a flexible process in the TSDuck repository to install the current TSDuck version in the current user's environment. This isn't immutable at all. However, the TSDuck repository is constantly evolving and cannot content immutable references.

> I agree. I use both Mac and Linux, and this allow me to have the same workflow in both environments... It's not perfect. I dislike a lot of things about Nix, but it does solve some problems well.

Well, we can say the same with Homebrew. It also works on Linux and can be used as single workflow in both environments. However, I would say that Homebrew is "native" in macOS and "alien" in Linux. From decades of experience, I am quite reluctant to frameworks which pretend to be univeral. There are excellent in their native environment, where and for which they were originally developed, and quite poor in other environments, where they cannot compete with native environments, and often create more problems than they solve. As a rule of thumb, I would recommend to always use the native tools in each environment. They are always the best. This is not only true for package managers, it also applies to compilers, code libraries, hypervisors, etc.

> I'll revert the test changes and run something simple.
> Is there a way to validate plugins built correctly and can be loaded? May be useful for SRT, etc.

Yes, that is exactly what the "test" section does in the [Homebre formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/t/tsduck.rb).

Basically, just run `tsp --version`, `tsp --list=input`, `tsp --list=output`, `tsp --list=packet` and check in their standard error that a required version or plugins are present.

> I did have a question about your branches. For the latest "stable" release, currently v3.43-4549, should I update the docs to point to the develop branch, as that seems to be tied to the latest stable release, or would you prefer I use explicit version

You can't update the doc for version 3.43. It is frozen and shipped with the released packages for that version.

The TSDuck repository contains the doc for the current commit in the repository. The site [tsduck.io](https://tsduck.io) contains online versions for the current state of the repository as well. The binary packages for version vX.Y-Z install the doc for that version. This means that a user can find the documentation for his own installed version on his own system. The online documentation describes the current state of the development.

I would recommend that the procedure (script, makefile, whatever) in the repository installs the current state of the repo. Provide an option to specify a precise official version, for which there is a tag in the repository. If you describe that in the documentation, the doc becomes independent of the version.

As a general rule, do not specify any specific version in the scripts in the repo. It should work on current state by default and accept any specific version from the user using a dedicated option (if you feel that this is a required feature). Of course, in the documentation, you may specify a given version as an example.

jcekstrom:
@lelegard I think it is ready.  I have tested the build on linux x86_64, ARM64, and macOS ARM.  Let me know if you have any other questions/concerns.

Issue #1701, opened 2026-04-11
scripts/setenv.sh: Support sourcing from zsh


piotrserafin:
Currently, `scripts/setenv.sh` uses bash-specific features such as `BASH_SOURCE[0]` to determine the script's directory and `${!varname}` for indirect variable expansion. This causes errors when attempting to source the script from `zsh`.

Specifically:
- `BASH_SOURCE[0]` evaluates to empty in `zsh`, breaking the resolution of `$ROOTDIR` and `$BINDIR`.
- `${!varname}` causes a bad substitution error in `zsh`.

### Proposed Solution
Update `scripts/setenv.sh` to correctly identify the script path in both `bash` and `zsh` environments, and use `eval` for indirect variable expansion so that it is compatible with both shells.

lelegard:
Fixed with PR #1703

PR #1703, opened 2026-04-11
scripts: support sourcing setenv.sh from zsh

piotrserafin:
#### Related issue (if any):
#1701

#### Affected components:
`scripts/setenv.sh`

This PR updates the `scripts/setenv.sh` utility so that it can be safely and reliably sourced from zsh as well as bash. The script previously relied on bash-specific features (`BASH_SOURCE[0]` and indirect variable expansion `${!varname}`), which triggered errors when executed in zsh (the default shell on macOS).
Specific changes include:
- Implementing shell-agnostic path resolution for `THIS_SCRIPT` using `$ZSH_VERSION` and `$BASH_VERSION`.
- Replacing the bash-only indirect expansion with a safe `eval` equivalent that works correctly across different shells.
- Fixing warnings raised by shellcheck

#### Tested Env:
- macOS 26.4
- Shells:
  - zsh 5.9 (arm64-apple-darwin25.0)
  - GNU bash, version 5.3.9(1)-release (aarch64-apple-darwin25.1.0)
  - GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25) (SH)

### DIFF
--- a/scripts/setenv.sh
+++ b/scripts/setenv.sh
@@ -16,7 +16,16 @@
 #
 #-----------------------------------------------------------------------------
 
-SCRIPT=$(basename "${BASH_SOURCE[0]}")
+if [ -n "${BASH_VERSION:-}" ]; then
+    THIS_SCRIPT="${BASH_SOURCE[0]}"
+elif [ -n "${ZSH_VERSION:-}" ]; then
+    # shellcheck disable=SC2296
+    THIS_SCRIPT="${(%):-%x}"
+else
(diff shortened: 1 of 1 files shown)


lelegard:
Hi Piotr,

Thanks for your contribution. I exclusively use bash on all systems (including Windows😉) and I lack practice of other shells.

BTW, the failures of the tests are due to your `tsduck-test` forked repo which is not up-to-date.

PR #1727, opened 2026-06-09
Nolnbcontrol

bap833:
#### Affected components:
tsp -I dektec

Some Dektec input channels don't support LNB control (like odd channels of the DTA-2128) and will have this error:
```
* dektec: using Dektec device 0, input channel 1 (DTA-2128 port 2)
* Error: dektec: error enabling Dektec LNB controller: DTAPI_E_NOT_SUPPORTED (DTAPI status 4119)
```
Whereas input channel 0 is working fine because it supports LNB control

```
tsdektec -av
DTAPI: 6.13.0.242, DtPcie: 3.6.4.398, DtPcieNw: -1.-1.-1.-1, Service: 5.2.8.141

* Device 0: DTA-2128 (Octal DVB-S2X Receiver)
  Physical ports: 8
  Channels: input: 8, output: 0
  Input 0: DTA-2128 port 1 (top), Demodulator (**LNB**, Automatic spectral inversion
      correction., Automatic symbol rate detection., Manual symbol rate., L-band
      950-2150MHz, Uni-directional input, Transparent mode, MPEG-2 transport
      stream, Demodulation, Adjustable roll-off factor, DVB-S reception, DVB-S2
      reception, DVB-S2 multiple input stream reception)
  Input 1: DTA-2128 port 2, Demodulator (Automatic spectral inversion
      correction., Automatic symbol rate detection., Manual symbol rate., L-band
      950-2150MHz, Uni-directional input, Transparent mode, MPEG-2 transport
      stream, Demodulation, Adjustable roll-off factor, DVB-S reception, DVB-S2
      reception, DVB-S2 multiple input stream reception)
   Input 2: [...]
```

It seems that dektec input plugin will always try to control the LNB when DVB-S or DVB-S2 modulations are selected.
The idea is to add an option to allow to disable LNB control when it's not needed

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,7 +7,12 @@ VERSION 3.45-4693 (May 2026)
 [IMP] Improvements on existing commands and plugins:
 
   * New options in existing commands and plugins:
-    - Option --no-lnb-control added in input plugin "dektec".
+    - Option --no-lnb-control in plugin "dektec" (input) to explicitly disable
+      LNB control on DVB-S/S2 receivers.
+
+  * Plugin "dektec" (input): on DVB-S/S2 receivers, LNB control is now
+    automatically skipped on input channels which do not support it instead of 
+    failing.
 
 -------------------------------------------------------------------------------
 
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,6 +2,15 @@ TSDUCK CHANGE LOG AND RELEASE NOTES
 
 -------------------------------------------------------------------------------
 
+VERSION 3.45-4693 (May 2026)
+
+[IMP] Improvements on existing commands and plugins:
+
+  * New options in existing commands and plugins:
+    - Option --no-lnb-control added in input plugin "dektec".
+
+-------------------------------------------------------------------------------
+
 VERSION 3.45-4684 (May 2026)
 
 [IMP] Improvements on existing commands and plugins:
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -14,6 +14,7 @@ Eric De Jong
 William del Strother
 Lola Delannoy
 Joey Ekstrom
+Baptiste Fave
 Gareth Fenn
 Martin Frétigné
 Thomas Göttgens
--- a/doc/user/plugins/dektec-input.adoc
+++ b/doc/user/plugins/dektec-input.adoc
@@ -228,11 +228,12 @@ Must be one of `ATSC-VSB`, `DAB`, `DVB-C2`, `DVB-S`, `DVB-S-QPSK` (same as `DVB-
 *--no-lnb-control*
 
 [.optdoc]
-DVB-S/S2 receivers: do not send any control command to the LNB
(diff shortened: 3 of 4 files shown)


lelegard:
Hi @bap833 

Thank you for reporting this.

Let me comment on the fix.

A command line option is a mean to give to the user a way to select between different possible actions and the choice between them is the responsibility of the user. Here, the choice is not the responsibility of the user. It's the capability of the channel. So, the right fix should be testing the capability of the channel (because it is possible, based on the command output) and apply LNB control only when it is supported by the channel. And maybe issue a verbose-level message when it is not.

bap833:
Hello @lelegard 

Thanks a lot for your quick answer

Indeed, I understand the logic and it would be nice to automatically disable LNB control if not supported to avoid the error.

I did not mention it earlier, but my idea behind the --no-lnb-control option was to cover also the scenario where you don't actually have an LNB to control, and you need to use a DC blocker to protect your device upstream. In that case this might be the responsability of the user to disable LNB control regardless of the capabilities of the channel.

I can implement the automatic disabling of LNB control if not supported, but should we also keep the --no-lnb-control option ?

lelegard:
> I did not mention it earlier, but my idea behind the --no-lnb-control option was to cover also the scenario where you don't actually have an LNB to control, and you need to use a DC blocker to protect your device upstream. 

I understand. These are two independent issues in that case.

The first one is a fix: don't try to send LNB control on a device that doesn't support the feature. The user shouldn't need to know which channel supports it and shouldn't need to specify an option to avoid something that is known to fail.

The second one is a feature: explicitly disable LNB control in a lab where sending those signals on the cable is a problem.

The two issues can be implemented in the same PR.

bap833:
I see 2 ways to fix it :

- Testing the capability of the channel (clean, but will not enable LNB if capability is not well reported by hardware. A different setup that was working well before could be failing because of this fix)
- Catching the DTAPI_E_NOT_SUPPORTED error, and skip LNB setup (less clean, but safer and simple)

Let me know which one you prefer

lelegard:
In that specific case, DTAPI is provided by Dektec for Dektec devices and I trust the adequacy between the two. So, the first solution is preferred.

I wouldn't come to the same conclusion with LinuxTV or DirectShow where the API, the drivers, and the hardware come from three different origins.

bap833:
The fix works well on my DTA-2128:

```
tsp -v -I dektec --device 0 --channel 0 --modulation DVB-S2 --frequency 1759500000 --symbol-rate 24000000 -O drop
* dektec: using Dektec device 0, input channel 0 (DTA-2128 port 1)
* dektec: unknown initial input bitrate
* dektec: new input bitrate: 23,732,255 b/s
* dektec: new input bitrate: 23,732,252 b/s
```

```
tsp -v -I dektec --device 0 --channel 1 --modulation DVB-S2 --frequency 1759500000 --symbol-rate 24000000 -O drop
* dektec: using Dektec device 0, input channel 1 (DTA-2128 port 2)
* dektec: input channel does not support LNB control, skipping it
* dektec: unknown initial input bitrate
* dektec: new input bitrate: 23,732,260 b/s
* dektec: new input bitrate: 23,732,249 b/s
```

PR #1745, opened 2026-08-25
Fix build with OpenSSL 4.0

WXbet:
#### Related issue (if any):
None.

#### Affected components:
TLS support of `libtscore` on Unix: `tsOpenSSL.h`, `tsTLSContextGuts.cpp`, `tsTLSCertificateGuts.cpp`.

Building TSDuck against OpenSSL 4.0 currently fails. Three incompatibilities:

1. `OPENSSL_atexit()` no longer exists. `tsOpenSSL.h` already maps it to `atexit()`, but only in the branch for versions below 3, which was added for LibreSSL and which OpenSSL 4 never reaches. Version 4 gets its own branch.

2. `SSL_set1_host()` and `SSL_add1_host()` are deprecated. The names to verify are now set through `X509_VERIFY_PARAM_set1_host()` and `X509_VERIFY_PARAM_add1_host()` on the verification parameters of the connection, which is what the SSL-level wrappers called internally.

3. `X509_get_subject_name()` returns a const pointer, so the subject name of a certificate can no longer be filled in place. It is built with `X509_NAME_new()` and passed to `X509_set_subject_name()`, which copies it.

Two points for your consideration:

For the second item, OpenSSL 4.0 recommends the new `SSL_set1_dnsname()` and `SSL_set1_ipaddr()` instead. Those exist only from 4.0 on, so using them needs a version switch, and `SSL_set1_dnsname()` validates the DNS name structure, which would reject a server name given as a literal IP address. Going through the verification parameters avoids both, needs no version switch and cannot change behaviour, since it is the same code path the wrappers used.

The certificate change merges two error messages into one, because the three calls are now a single `&&` expression with a `X509_NAME_free()` after it. Keeping them apart would need a free on every error path.

### DIFF
--- a/src/libtscore/tls/unix/tsOpenSSL.h
+++ b/src/libtscore/tls/unix/tsOpenSSL.h
@@ -44,6 +44,11 @@
         // emulates OpenSSL v3, except that OPENSSL_atexit is not available.
         #define OPENSSL_atexit atexit
     #endif
+    #if OPENSSL_VERSION_MAJOR >= 4
+        // OpenSSL v4 removed OPENSSL_atexit().
+        #include <cstdlib>
+        #define OPENSSL_atexit atexit
+    #endif
     #include "tsAfterStandardHeaders.h"
 
 #else
--- a/src/libtscore/tls/unix/tsTLSCertificateGuts.cpp
+++ b/src/libtscore/tls/unix/tsTLSCertificateGuts.cpp
@@ -220,13 +220,20 @@ bool ts::TLSCertificate::createEphemeralCertificate(size_t rsa_bits)
         }
 
         // Set subject name (CN) and issuer to host name (this is a self-signed certificate).
-        X509_NAME* subject_name = X509_get_subject_name(x509); // internal pointer inside x509
-        if (!X509_NAME_add_entry_by_txt(subject_name, "CN", MBSTRING_ASC, reinterpret_cast<const unsigned char*>(Subject().toUTF8().c_str()), -1, -1, 0)) {
-            report().error(u"error setting subject of certificate");
+        // X509_get_subject_name() returns a const pointer since OpenSSL 4.0, so the
+        // name is built separately and handed to the certificate, which copies it.
+        X509_NAME* const subject_name = X509_NAME_new();
+        if (subject_name == nullptr) {
+            report().error(u"error allocating subject of certificate");
             break;
         }
-        if (!X509_set_issuer_name(x509, subject_name)) {
-            report().error(u"error setting issuer of certificate");
+        const bool subject_ok =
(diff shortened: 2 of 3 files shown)


lelegard:
Hi @WXbet 

Thanks for doing the analysis. Did you try to run the unitary tests? In a native build, this is done using `make test` but with a cross-build you have to run the `utest` program, as built with the rest of TSDuck.

On my side, with a natively recompiled OpenSSL 4.0.2 on Ubuntu 26.04 on x64, it fails.

What works:

- Compilation with OpenSSL 3.x and 4.0.2.
- Run with OpenSSL 3.x.

What doesn't work:

- Run with OpenSSL 4.0.2.

~~~
tsversion --version=tls
OpenSSL 4.0.2 (CPUINFO: OPENSSL_ia32cap=0xfefa3203478bffff:0x00000000208c2569:0x0000000030000400:0x0000000000000000:0x0000000000000000)

utest -d -t reactivetlstest::client
== Running test ReactiveTLSTest::Client
TextClient: connecting to tsduck.io (87.98.154.146:443)
* Error: OpenSSL error: SSL_read
* Error: 005DEA684B7B0000:error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2497:
TextClient::handleTCPConnected: error code: -1
* Error: OpenSSL error: SSL_do_handshake
* Error: OpenSSL error: SSL_write
TextClient::handleTextLine: end of response
* Error: OpenSSL error: SSL_shutdown
* Error: OpenSSL error: SSL_shutdown
* Error: OpenSSL error: SSL_shutdown
... infinitely loop on last error line ...
~~~

The failing test is a TLS client connecting to https://tsduck.io/, just reading the returned HTTP headers and HTML text.

This is a "reactive" test, meaning using non-blocking I/O and a reactor pattern (on Linux, this an event dispatcher based on epoll). This means that we don't let OpenSSL access the network. We do the non-blocking network access in TSDuck code and use OpenSSL as a state machine working on pure data, through `BIO`.

WXbet:
Let me check it.

WXbet:
With a private OpenSSL build there is no trust store, so certificate verification always fails. Please try:

`SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt utest -d -t reactivetlstest::client`

With that, the test passes here against OpenSSL 4+.

```
== Running test ReactiveTLSTest::Client
TextClient: connecting to tsduck.io (87.98.154.146:443)
TextClient::handleTCPConnected: error code: 0
TextClient::handleTextLine: "HTTP/1.1 200 OK", error code: 0
...
TextClient::handleTextLine: end of response
TextClient::handleTCPClosed

OK (1 tests, 16 assertions)
```

So, with a native built tsduck you simply run `SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt make test` and it should be fine.

lelegard:
My bad, you are right. I just built and run the test, without investigating the OpenSSL configuration. I will merge it.

